PR #6, opened 2018-10-08
Assert-free delay implemtation

gbip:
Assert-free delay implemtation from the one by @jamesmunns in the stm32f103xx-hal crate.

The [stm32f103 implementation](https://github.com/japaric/stm32f103xx-hal/blob/master/src/delay.rs#L48).

### DIFF
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -47,17 +47,29 @@ impl DelayMs<u8> for Delay {
 
 impl DelayUs<u32> for Delay {
     fn delay_us(&mut self, us: u32) {
-        let rvr = us * (self.clocks.hclk().0 / 8_000_000);
+        // The RVR register is 24 bits wide, as SysTick is based on a 24 bit counter
+        const MAX_RVR: u32 = (1 << 24);
 
-        assert!(rvr < (1 << 24));
+        let mut total_rvr = us * (self.clocks.hclk().0 / 8_000_000);
 
-        self.syst.set_reload(rvr);
-        self.syst.clear_current();
-        self.syst.enable_counter();
+        while total_rvr != 0 {
+            let current_rvr = if total_rvr < MAX_RVR {
+                total_rvr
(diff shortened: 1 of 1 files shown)


jamesmunns:
Actually looking at this, I think I might have made an off-by-one error here.

src/delay.rs
```diff
@@ -47,17 +47,29 @@ impl DelayMs<u8> for Delay {
...
+        let mut total_rvr = us * (self.clocks.hclk().0 / 8_000_000);
 
-        self.syst.set_reload(rvr);
-        self.syst.clear_current();
-        self.syst.enable_counter();
+        while total_rvr != 0 {
+            let current_rvr = if total_rvr < MAX_RVR {
+                total_rvr
+            } else {
+                MAX_RVR
```
jamesmunns: @therealprof I think this should actually be `MAX_RVR - 1` (or `MAX_RVR` should be `(1 << 24) - 1`).

therealprof: I think you're right.

gbip: Shouldn't `RVR` be 2^24 ?

therealprof: The maximum value for RVR is 2^24 - 1. 2^24 would be treated as 0 which would essentially disable the SysTick counter since it triggers and reloads when going from 1 to 0.

jamesmunns: I would suggest replacing MAX_RVR's value with `0x00FFFFFF`, as that makes it clear it is a full 24 bit register.

From the ARM manual:

> The SysTick Reload Value register supports values between 1 and 0x00FFFFFF.

jamesmunns: Also when you have an updated version, could you please re-ping me? I know a couple crates have already copy/pasted this code, and I'd like to open issues.

gbip: My bad you are right !

PR #7, opened 2018-10-08
Fixed the maximum RVR value when creating a delay.

gbip:
See the discussion on #6

### DIFF
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -48,7 +48,7 @@ impl DelayMs<u8> for Delay {
 impl DelayUs<u32> for Delay {
     fn delay_us(&mut self, us: u32) {
         // The RVR register is 24 bits wide, as SysTick is based on a 24 bit counter
-        const MAX_RVR: u32 = (1 << 24);
+        const MAX_RVR: u32 = (1 << 24) - 1;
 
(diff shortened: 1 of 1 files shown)


src/delay.rs
```diff
@@ -47,13 +47,13 @@ impl DelayMs<u8> for Delay {
...
-        // The RVR register is 24 bits wide, as SysTick is based on a 24 bit counter
-        const MAX_RVR: u32 = (1 << 24) - 1;
+        // The SysTick Reload Value register supports values between 1 and 0x00FFFFFF.
+        const MAX_RVR: u32 =  0x00FF_FFFF;
 
         let mut total_rvr = us * (self.clocks.hclk().0 / 8_000_000);
 
         while total_rvr != 0 {
-            let current_rvr = if total_rvr < MAX_RVR {
+            let current_rvr = if total_rvr <= MAX_RVR {
```
therealprof: This should not actually make any difference. ;)

jamesmunns: Oh right. Code is hard :)

gbip: Yep I followed @jamesmunns without thinking. I thought it would enable some kind of optimization but even godbolt [disagrees](https://rust.godbolt.org/).

PR #8, opened 2018-10-08
Add a config struct to control uart async data framing

timbod7:
This is proposed for comments. Assuming it looks ok, I will document and implement onthe other usarts.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -84,28 +84,34 @@ pub mod config {
             }     
         }
 
-        pub fn set_parity_none(&mut self) {
-            self.parity = Parity::ParityNone
+        pub fn parity_none(mut self) -> Self {
+            self.parity = Parity::ParityNone;
+            self
         }
 
-        pub fn set_parity_even(&mut self) {
-            self.parity = Parity::ParityEven
+        pub fn parity_even(mut self) -> Self {
+            self.parity = Parity::ParityEven;
+            self
         }
 
-        pub fn set_parity_odd(&mut self) {
-            self.parity = Parity::ParityOdd
+        pub fn parity_odd(mut self) -> Self {
+            self.parity = Parity::ParityOdd;
+            self
         }
 
-        pub fn set_wordlength_8(&mut self) {
+        pub fn wordlength_8(mut self) -> Self {
             self.wordlength = WordLength::DataBits8;
+            self
         }
 
-        pub fn set_wordlength_9(&mut self) {
+        pub fn wordlength_9(mut self) -> Self {
             self.wordlength = WordLength::DataBits9;
+            self
         }
 
-        pub fn set_stopbits(&mut self, stopbits: StopBits) {
+        pub fn stopbits(mut self, stopbits: StopBits) -> Self {
             self.stopbits = stopbits;
+            self
         }
     }
 
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -12,6 +12,8 @@ use stm32::{RCC, UART4, UART5, USART1, USART2, USART3, USART6};
 #[cfg(feature = "stm32f429")]
 use stm32::{RCC, UART4, UART5, UART7, UART8, USART1, USART2, USART3, USART6};
 
+use stm32::usart6::cr2::STOPW;
+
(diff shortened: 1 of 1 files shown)


timbod7:
> The approach should be independent of any MCU specific flags

I didn't see the need for this given that the _construction_ of the serial interface is entirely MCU dependent (ie you specify the actual USART and pins to use).

Are you suggesting I should make a local copy of the STOPW enum, and them map back to the MCU specific one?

> There should be some builder functions to make up the configuration

I'm new to rust, and am largely unaware of preferred coding patterns. Given that the internal implementation is unlikely to change, what's the reason to prefer a builder over a simple struct with a corresponding default value?

timbod7:
Thanks for the answers - I'll follow up with another commit.

> You chose to use stm32::usart6::cr2::STOPW to initialise USART2 

Is that wrong? I think the MCU crate generated by svd2rust only generates the types for usart6, and then reuses them for the other compatible USARTs.

therealprof:
> Is that wrong? I think the MCU crate generated by svd2rust only generates the types for usart6, and then reuses them for the other compatible USARTs.

Actually you're right, I was reading that as capital `USART6`. However the UARTs have their set of registers produced as `uart4` and those are not compatible with the `usart6` registers and values, so my point still stands. ;)

timbod7:
Hopefully close to what you had in mind?

therealprof:
Looks good on first glance, except that it's not using the builder pattern. ;)

In your approach you'd do something like:
```
let mut config = Config::new(115_200);
config.set_parity_even();
config.set_wordlength_9();
...
let mut serial = usart2(..., config);
```

OTOH the builder pattern returns takes ownership of the structure and then returns the modified structure back to the caller, so you'd end up with:
```
let config = Config::new(115_200).set_parity_even(). set_wordlength_9();
...
let mut serial = usart2(..., config);
```
or you could just do it in a single `let mut serial` line, too:
```
let mut serial = usart2(..., Config::new(115_200).set_parity_even(). set_wordlength_9());
```

I'd probably get rid of the `set_`, too, for brevity sake.

I'll have to inspect the rest of it a bit closer later and give it a test run.

PS: What are we going to do about the other U(S)ARTs? I'm afraid we'll either have to multiply the changes in or we'll have to macrofy it first.

timbod7:
> What are we going to do about the other U(S)ARTs? I'm afraid we'll either have to multiply the changes 

I assumed that `serial.rs` wasn't using macros for a reason.

Once the changes on USART2 looked ok, I intended to multiply the changes.

therealprof:
> I assumed that serial.rs wasn't using macros for a reason.

My laziness. ;)

timbod7:
I thought that the new method that took all non defaultable parameters was
preferable. But happy to change it to a default instance with an arbitrary
baud rate.

therealprof:
Whether to supply parameters to new or not is a preference thing. Providing a `Default` implementation is almost always a good thing and there's a clippy lint happily complaining if you have a `new` method but no `default` method.

timbod7:
On switching back to a default trait rather than a `new()` method it seems a little harder to use... 

This:

```
        let config : serial::config::Config = Default::default()
            .baudrate(100_000.bps())
            .parity_even()
            .wordlength_9()
            .stopbits(serial::config::StopBits::STOP2);
```
doesn't build:

```
  --> src/main.rs:60:47
   |
60 |         let config : serial::config::Config = Default::default()
   |                                               ^^^^^^^^^^^^^^^^^^ cannot infer type for `Self`
   |
   = note: type must be known at this point
```

I seem to need to write:

```
        let config : serial::config::Config = Default::default();
        let config = config
            .baudrate(100_000.bps())
            .parity_even()
            .wordlength_9()
            .stopbits(serial::config::StopBits::STOP2);
```

is there a better way?

timbod7:
I was still expecting to duplicate to the other USARTs....

therealprof:
Try:
```
let config = serial::config::Config::default();
```

timbod7:
BTW. Thanks for your guidance along the way. This is pretty much my first rust code.

therealprof:
@timbod7 Do that in another PR and earn yourself a Hacktoberfest shirt. ;)

PR #16, opened 2018-10-31
Add support for the Timer abstraction

aurelj:


src/rcc.rs
```diff
@@ -275,6 +283,16 @@ impl Clocks {
         self.pclk2
     }
 
+    /// Returns the prescaler of the APB1
+    pub fn ppre1(&self) -> u8 {
+        self.ppre1
+    }
+
+    /// Returns the prescaler of the APB1
```
therealprof:
```suggestion
    /// Returns the prescaler of the APB2
```

aurelj:
Thanks for the review.
Branch updated with the typo fixed and STM32F412 support added.

PR #19, opened 2018-10-31
Add timer module with initial support for STM32F412.

benkard:
This goes together with https://github.com/therealprof/stm32f4xx-hal/pull/17.

Most of the implementation is taken from stm32f30x_hal.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ version = "0.2.2"
 cortex-m = "0.5.8"
 cortex-m-rt = "0.6.5"
 nb = "0.1.1"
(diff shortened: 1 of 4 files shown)


benkard:
https://github.com/therealprof/stm32f4xx-hal/pull/16 seems to do the same thing, maybe with some slight differences.  I wasn't aware of that PR when I made this branch.  Feel free to reject my PR if the existing one is better.

therealprof:
Yeah, plenty of PRs raised today. 👍 I think I'll take the other PR on the merit of both being first and looking a bit easier on the eye for me.

I'm sure there's still a lot todo so feel free to chime in any time.

PR #20, opened 2018-11-04
rcc: keep the cache bits in FLASH_ACR intact

birkenfeld:
This allows configuring the cache bits independent of clock freeze.

An alternative would be to switch the caches on in here?


therealprof:
@birkenfeld I think we should add functionality to configure the caches. But we should certainly keep whatever is in the registers already.

PR #22, opened 2018-11-05
Allow compiling without any feature selected

aurelj:
This fixes `cargo publish`.

The last patch may not be the prettiest but it makes some sens anyway.


therealprof:
I wonder if there's a way to indicate that a user of this should select a MCU for this to be useful without breaking publish.

PR #24, opened 2018-11-07
SPI: Add support for STM32F4{01,07,12,29}

psycowithespn:
I based this off of spi.rs in stm32f103xx-hal. Verified on an F407 with a device and F429 by logic analyzer / loopback.

The SPI4/5/6 block should be left commented until https://github.com/adamgreig/stm32-rs/pull/83 is fully released (next version) and this crate updates to the new version.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -287,7 +287,6 @@ macro_rules! hal {
                     clocks: Clocks
                 ) -> Self
                 where PINS: Pins<$SPIX> {
-
                     // NOTE(unsafe) This executes only during initialisation
                     let rcc = unsafe { &(*RCC::ptr()) };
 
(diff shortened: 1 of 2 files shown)


src/spi.rs
```diff
@@ -0,0 +1,422 @@
...
+            impl<PINS> Spi<$SPIX, PINS> {
+                pub fn $spiX(
+                    spi: $SPIX,
+                    pins: PINS,
+                    mode: Mode,
+                    freq: Hertz,
+                    clocks: Clocks
+                ) -> Self
+                where PINS: Pins<$SPIX> {
+
```
therealprof:
```suggestion
```

src/spi.rs
```diff
@@ -0,0 +1,422 @@
...
+    }
+}
+
+hal! {
+    SPI1: (spi1, apb2enr, spi1en, pclk2),
+    SPI2: (spi2, apb1enr, spi2en, pclk1),
+    SPI3: (spi3, apb1enr, spi3en, pclk1),
+}
+
+// TODO(psycowithespn) Enable all of these on the next version of stm32-rs.
```
therealprof: Please remove this block for now. I don't want random todos hanging around in the code.

therealprof:
Thanks for the PR, LGTM except for the todo block and the one nit...

psycowithespn:
Applied both requested changes. 

I can squash it if you'd like.

PR #27, opened 2018-11-15
Added STOP to end of Write, per I2C spec.  

osterwood:
To do this, created send private 'send_without_stop' and 'send_stop' methods and changed Write & WriteRead to use it.  This is modeled after https://github.com/japaric/stm32f103xx-hal/blob/master/src/i2c.rs

Sorry for the strange looking diff -- the Write and WriteRead methods didn't move, but the large move of code from within Write to inside of the i2c1 impl makes it look like they did.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -436,7 +436,7 @@ macro_rules! gpio {
                         let offset = $i;
                         unsafe {
                             &(*$GPIOX::ptr()).otyper.modify(|r, w| {
-                                w.bits(r.bits() & (1 << offset))
+                                w.bits(r.bits() | (1 << offset))
                          })};
 
                         self
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -1,254 +1,496 @@
-use stm32::{I2C1, RCC};
+use stm32::{I2C1, I2C2};
+use stm32::DWT;
+
+// use afio::MAPR;
 
 use hal::blocking::i2c::{Read, Write, WriteRead};
+use nb::{Error as NbError, Result as NbResult};
+use nb::Error::{Other, WouldBlock};
 
-use gpio::gpiob::{PB6, PB7, PB8, PB9};
+use gpio::gpiob::{PB10, PB11, PB6, PB7, PB8, PB9};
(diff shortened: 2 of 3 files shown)


therealprof:
Thanks. Looks good on first glance but I'd like to test it which I can't at the moment.

osterwood:
Interesting.  I'm running an Nucleo-F401RE with some external I2C devices and it works fine at 400 kHz (after syncing with your recent changes to fix the I2C operating frequency).  I'm testing with an SSD1306 OLED and TLC59208 LED drivers.

therealprof:
@osterwood Interesting, I'm also testing with an SSD1306. At which `sysclk` are you running?

therealprof:
I'm using the `i2c_hal_ssd1306alphabeter` example from git@github.com:therealprof/stm32f429i-disc.git if you're interested.

osterwood:
Sorry - I didn't realize that my recent pushes to the fork would be included in this PR (guess I should have branched master before creating the PR).  This was not intended.  The commit is large and full of macros - I agree that it's ugly.

I like the approach taken in #37, but for now want the I2C library to handle retry / not permanently waiting for ACKs.  So, I will keep using my crazy fork for now.  Definitely will be watching #37 and other changes to this HAL and expect to come back to it in the future.

To answer your question from a few days ago, running at 40 MHz right now.  Test code is based on this: https://github.com/jamwaffles/ssd1306/blob/master/examples/terminal_i2c.rs but ported to his HAL.   For now, just close this PR.  I'll test usage of the TLC59208 against #37 and let you know how that goes (that was the device which needed STOPs to work).

therealprof:
I'd be happy to take your STOP patch if it continues to work for me. The retry patch is a bit too much magic for my taste; my preference here would be to use the non-blocking I2C impl instead and handle it in the user code. The SSD1306 example is also from me BTW and a variant of the mentioned `i2c_hal_ssd1306alphabeter`. 😉

TeXitoi:
Any update on this? or can we close it?

osterwood:
Oh for sure just close it.

PR #35, opened 2018-11-19
rcc: stm32f411 min sysclk is 12.5 MHz like most stm32f4

aurelj:


therealprof:
Where are those numbers coming from? If you use PLL, the minimum value you can configure is 24MHz; if you're using HSE the minimum seems to be 4MHz. (all values for STM32F411)

aurelj:
The minimum value is 24 MHz only for stm32f401...

RM0383 6.3.2 RCC_PLLCFGR
"ensure that the VCO output frequency is between 100 and 432 MHz"
sysclk is VCO output divided by PLLP and max PLLP value is 8.
So min sysclk is 100 / 8.

aurelj:
Oh, and yes, those values are only relevant in the PLL code path. You can indeed go down to 4 MHz using HSE.

therealprof:
@aurelj CubeMX tells me that after applying PLLP the frequency shall be no less than 24MHz (on the STM32F411).

aurelj:
Maybe CubeMX has a single safe value that apply to the whole stm32f4 family including the stm32f401 ?

Looking at the reference manuals and datasheet of the whole F4 family, I see no such 24 MHz limitation except for the stm32f401.

I have no stm32f4 except a F401 so I can't really test, but anybody should be able to test this using the PLL with the following parameters:
- PLLSRC: 0
- PLLM: 16
- PLLN: 100
- PLLP: 8

therealprof:
> Maybe CubeMX has a single safe value that apply to the whole stm32f4 family including the stm32f401 ?

I doubt it. CubeMX is pretty smart and will adjust stuff to the processor at hand. The 401 for instance does not allow for external I2S clock and has a lower maximum speed compared to the 411. The clocks for a 469 on the other hand look vastly different...

therealprof:
I'm not saying CubeMX is necessarily right but it certainly has the benefit of the doubt over non-specified values here... ;) Probably there's some AN that describes the minimum value somewhere.

aurelj:
Not sure why you say "non-specified" values ? The datasheet makes it pretty clear that the minimum VCO output frequency is 100 MHz (except for the F401 where the datasheet specifies a minimum of 192 MHz).

But I would be glad to hear about an AN or any other doc specifying otherwise.

adamgreig:
The datasheet for the STM32F411 explicitly gives a minimum for the PLL output:

image

therealprof:
@adamgreig Thanks a lot.

aurelj:
Oh, it seems I missed that one. I still find it a bit strange that the reference manual explicitly specify a min and max for PLLM out and PLLN out, but only specifies a max for PLLP out.

But anyway, looking at all the stm32f4 datasheets, they all specify a minimum of 24 MHz except the stm32f446 witch has a 12.5 MHz minimum...
I will update this PR to follow this datasheet minimum.

Thanks @adamgreig for pointing this out.

therealprof:
Hm, in case of the STM32F446 CubeMX would be wrong then because it insists on a minimum of 24MHz still.

PR #39, opened 2018-11-21
OTy is 1 for open drain, not 0

psycowithespn:
Found my I2C issue. 😄

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -436,7 +436,7 @@ macro_rules! gpio {
                         let offset = $i;
                         unsafe {
                             &(*$GPIOX::ptr()).otyper.modify(|r, w| {
-                                w.bits(r.bits() & (1 << offset))
+                                w.bits(r.bits() | (1 << offset))
                          })};
 
                         self


aurelj:
Note that this is exactly this kind of bugs that #36 would avoid or at least make much more obvious.

therealprof:
@psycowithespn lol, I just found the exact same issue 30 minutes ago. :-D

therealprof:
@aurelj Maybe, maybe not. Excessive macro causes similar problems, just in other places since it is easy to mix up the 7th and 8th parameter of 11 and endless repetition is also prone to cut and paste errors which we're still finding every now and then.

osterwood:
Fixed my issue too!  Had one I2C device that would not ACK prior to this change -- guess it was not happy with the sharp rising edge.  

**Prior to change : I2C pins acting as push / pull -- no ACK from device**

scr41

**After change : I2C in open-drain mode -- good ACK**

scr43

psycowithespn:
> guess it was not happy with the sharp rising edge.

I2C requires open-drain on everything. In push-pull, the MCU is actively driving the data line and the device is unable to pull it to ground to signal an ACK. The external pull up resistors are the only thing that should drive the lines high.

osterwood:
You're right that it was the mis-configuration (lack of open-drain) causing the issue with this device and not the ringing on the rising edges.  But very oddly -- several other I2C devices on the bus were working fine and ACKing all data.  Somehow they were able to pull the data line low even with the MCU configured improperly.

PR #41, opened 2018-11-26
Update the GPIO mode register after setting AF

psycowithespn:
This change prevents a small blip on the pin when changing to an alternate function. In my case, this was causing an I2C device to not respond. It was jumping to AF0 first, then AF4 a few cycles after.

The following was used to initialize the I2C pins.

```rust
let scl = gpiob.pb6.into_alternate_af4().set_open_drain();
let sda = gpiob.pb9.into_alternate_af4().set_open_drain();
```

Here is evidence of the small blip from before the change. After doesn't have this

pins blip

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -152,9 +152,6 @@ macro_rules! gpio {
                 let offset = 2 * index;
                 let offset2 = 4 * index;
                 unsafe {
-                    &(*$GPIOX::ptr()).moder.modify(|r, w| {
-                        w.bits((r.bits() & !(0b11 << offset)) | (0b10 << offset))
-                    });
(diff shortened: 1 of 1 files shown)


psycowithespn:
Good point. Changes made to all `moder` usages. The open-drain output mode had the same blip coming from reset since it is momentarily push-pull. 

The open-drain mode also defaults to output 0, so even if the user immediately does a `set_high`, there will be a blip. Should `into_open_drain_output` force an output of 1 to prevent this? Should `into_push_pull_output` also have a default, probably 0, to be consistent?

therealprof:
@psycowithespn Setting sensible default levels suitable for each mode before enabling it is probably a good idea.

PR #42, opened 2018-11-28
Added support for serial port idle interrupts

timbod7:


therealprof:
Looks fine to me but I'm curious about the usecase for this...

timbod7:
I'm doing some experimentation with drone firmware, and want to receive the SBUS protocol, which has a 25 byte packet. To detect the packet start, I need to wait for the line to be idle, then wait for the start byte (0xF0).

timbod7:
Thanks!

PR #43, opened 2018-11-29
Add support for all existing stm32f4 devices

aurelj:
This patch series adds support for every existing stm32f4 devices. It is not complete yet. Serial and I2C still needs to be done. Moreover, it requires a new release of the stm32f4 crates for some devices.

Anyway, you may still want to merge the patch series, except for the last patch, which is the one actually enabling all those new devices.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -514,6 +514,7 @@ macro_rules! gpio {
     }
 }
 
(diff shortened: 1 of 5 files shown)


aurelj:
Oh, indeed, it was probably not a good idea to include last patch in this pull request.

It is removed now, so this PR should be mergable.

Patch to actually enable all devices is still available here: https://github.com/aurelj/stm32f4xx-hal/commit/f92579b

PR #46, opened 2018-12-12
enable support for all existing stm32f4 devices

aurelj:
This requires a new release of stm32f4 first.


therealprof:
@aurelj The update is out now and the crate has been updated. 🎉

aurelj:
This should be OK to merge after #50 is pulled

PR #47, opened 2018-12-13
Add compile time barrier to make sure a device is chosen.

MabezDev:
This is something I just added to the l4 hal. I found when compiling, If I forgot to add the features I would get confusing error messages about the hal missing modules etc.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,12 +3,42 @@
 
 #[cfg(not(any(
     feature = "stm32f401",
+    feature = "stm32f405",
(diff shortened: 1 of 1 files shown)


MabezDev:
It's definitely possible with a build.rs script, but it does seem a bit overkill, as the other feature gates will need to be updated when new devices are added anyway.

> I'm going to defer this and request to add the new devices too when those PRs have landed

Sure let me know when they are merged and I will update the features on this PR.

PR #49, opened 2018-12-18
Add in serial Read and Write hal implementations for Serial

psycowithespn:
This change came about in my playing with `cortex-m-rtfm` and serial. What I discovered is that once you have the Tx and Rx structs, you can no longer mask and unmask the interrupts of the usart. This breaks common patterns like unmasking `TxE` to send data, then masking it when done transmitting.

This seems to be the more intuitive way as opposed to adding an `impl` of listen and unlisten to Tx and Rx.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -364,6 +364,17 @@ macro_rules! halUsart {
                 }
             }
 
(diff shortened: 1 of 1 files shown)


therealprof:
I agree, makes total sense to me.

PR #50, opened 2018-12-18
update stm32f4 to v0.5.0

aurelj:


aurelj:
Well, yes, it would be nice to be a collaborator.
I guess this would mean that I could merge PR myself. I would be OK to merge simple PR, but I would probably prefer to let you review more complex or controversial PR as well before merging.

therealprof:
@aurelj Absolutely. Whatever you feel up to comfortably. We still need a lot of cleanup and minor stuff to be done, like adding proper documentation, changelogs and tags.

aurelj:
Great ! I've merged this one as first test.
I guess the preferred style of merge is to crate a merge commit (rather than rebase and merge) as that's what you usually do.
I will look at the "enable all devices" PR tomorrow.

therealprof:
I usually use what's more convenient at the time. I'm actually more of a rebase kind of guy as that keeps the history clean.

aurelj:
OK, I also prefer rebase, so I will use this for future merge, when it make sense.

PR #53, opened 2018-12-21
serial: simplify baud rate divisor computation, fix fraction overflow

schodet:
The formula to compute USARTDIV (when OVER8 is 0) is:

    USARTDIV = round (f_CK / (16 * baudrate))

As USARTDIV is a fixed point number with 4 bits fractionnal part (when
OVER8 is 0), this simplifies to:

    USARTDIV_reg = round (f_CK / baudrate)

The original code was computing the rounded fractionnal part which could
overflow. In this case, it could cause an error of 1 when the mantisa is
odd.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -964,12 +964,9 @@ macro_rules! halUsartImpl {
                     rcc.$apbXenr.modify(|_, w| w.$usartXen().set_bit());
 
                     // Calculate correct baudrate divisor on the fly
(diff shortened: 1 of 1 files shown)


schodet:
Tested on a F407 discovery, using `serial_echo` example from `stm32f407g-disc` and a USB serial adaptater.

aurelj:
Indeed, this looks absolutely correct.
Thanks for the PR.

schodet:
Thanks.

PR #57, opened 2019-01-09
RCC configuration refactoring

albru123:
RCC configuration has been separated into methods and cleaned up a bit
PLL configuration has been extracted into separate entity
PLLQ value is now being set, so 48 MHz can be generated whenever
possible
IDEA folder (.idea) has also been added to .gitignore in the process

There might be some changes to code style, as IDEA has its own set of (not always great) style guidelines for code auto-formatting. I have nothing against changing it back if needed.

The code has been tested to work on real hardware by @therealprof, but might need a closer look.

This is my first PR, so I'd really like to hear your opinion on the code and its quality, possible improvements and mistakes I might have made.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 **/*.orig
 **/*.rs.bk
 Cargo.lock
+.idea/
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -1,6 +1,8 @@
-use core::cmp::min;
+use core::cmp::{max, min};
 
 use crate::stm32::{FLASH, RCC};
+use crate::stm32::rcc::pllcfgr::PLLSRCW;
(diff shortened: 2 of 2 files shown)


aurelj:
One thing that could be improved is making your changes in several smaller commits with one simple self contained change per commit. This would make it easier to review your changes. (I'm not asking you to do this right now, this can be quite painful to do afterwise)

Other than that, you can also have a look at #58 (Not saying my code is better, I just did it differently, which might give you some other ideas).

albru123:
@aurelj The changes were separate commits at first, but then I made some commits with WIP code to synchronize it between my machines, so I decided to squash all the changes to avoid confusion.

albru123:
Since the ideas I presented in this PR more or less made it into the HAL, I'll close this.

PR #58, opened 2019-01-09
Rcc improvements

aurelj:
- support for HSE
- support for setting AHB, APB1 and APB2 frequency
- code cleanup

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -15,6 +15,7 @@ impl RccExt for RCC {
     fn constrain(self) -> Rcc {
         Rcc {
             cfgr: CFGR {
+                hse: None,
                 hclk: None,
                 pclk1: None,
                 pclk2: None,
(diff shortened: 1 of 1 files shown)


aurelj:
Not at all.
But I just noticed the PR from @albru123... Not sure what to do now...

Oh, and the failure is due to the PPRE1 vs PPRE2 type alias. Locally I have some patches on top of this using #![feature(type_alias_enum_variants)] which fixes this (but is not supported on stable for now). But it would be quite easy (but uglier) to go back from PPRE* variants to plain bits for now, to avoid this issue.

therealprof:
#57 looks large but it's mostly refactoring and adding a few small bits around `PLLQ`. I think yours is more polished and we can easily add the other bits...

aurelj:
OK.

Issue fixed by using bits instead of variants for PPRE*, without changing anything else so that it is trivial later on to switch back to variants (when #![feature(type_alias_enum_variants)] will be available in stable).

albru123:
Looks better than mine (#57) for sure, but I would probably do some refactoring, so `freeze()` isn't that long.
I can _try_ do it once it's merged if you want.

PR #65, opened 2019-01-20
Add accessor for the uart status register (sr)

timbod7:


timbod7:
Hmmm. This builds with:

```
cargo build --target thumbv7em-none-eabi --features stm32f405,rt
```
but not with:
```
cargo build --target thumbv7em-none-eabi --features stm32f412,rt
```

Do you normally test with each feature flag?

timbod7:
> This crate is supposed to be a Hardware Abstraction Layer...

The HAL traits are `embedded_hal::serial::Read` and  `embedded_hal::serial::Write`, pretty much everything else in this file is to support implementing those. The issue with the current approach is that, if you want to use those traits, you lose access to all of the underlying non-abstract features (like interrupts, status etc). Hence the Serial type already exposes non-abstract features like interrupts (via `listen` and `unlisten`). Added access to the status register seemed to fit this pattern. 

But, I'd be fine with adding individual flags for is_idle(), is_rxne(), is_txe() if that approach is preferred.

timbod7:
PR updated. Access to 3 status bits is all I need (corresponding to the 3 event types). The implementation is much simpler.

aurelj:
Now this looks good to me.
I would personally prefer function names that are not specific to stm32 MCU, such as `is_data_available()` instead of `is_rxne()` and `is_space_available()` instead of `is_txe()`, but I would like to here other opinions about this.

therealprof:
@aurelj I agree about the naming. I would even go so far as to try to generalise this and propose a new trait for `embedded-hal`.

timbod7:
> I would personally prefer function names that are not specific to stm32 MCU

> I agree about the naming

The names I chose correspond to the existing event enum:

```
/// Interrupt event
pub enum Event {
    /// New data has been received
    Rxne,
    /// New data can be sent
    Txe,
    /// Idle line state detected
    Idle,
}
```

I think it makes sense to be consistent with this as the accessor functions are for discovering which event has occurred.

Are you suggesting that the enum should change also?

aurelj:
> I think it makes sense to be consistent with this as the accessor functions are for discovering which event has occurred.

It indeed makes sense.

> Are you suggesting that the enum should change also?

I would say so.

timbod7:
My preference would be to keep the even names consistent with the names of the underlying stm interrupts. I think this will reduce confusion for users - and it seems unlikely that this library will be used for non stm processors in the short term...

 But this is not my project - I'm happy to change to whatever names @therealprof  prefers.

timbod7:
One comment. The title of this library is "A Rust embedded-hal HAL for all MCUs in the STM32 F4 family". Given that, it seems preferable (and often unavoidable) to have STM32 specific naming within it, as anyone using this library should be familiar with STM processors.

Abstracted functions, names types and traits belong in the higher level  https://github.com/rust-embedded/embedded-hal project.

therealprof:
@timbod7 I haven't checked in detail whether it is feasible to use more general names. As I said, users of other hardware might have similar needs and thus if there's a chance to turn it into a more general trait. Having it in form of a trait which can be implemented has the big advantage that this can be used in universal drivers which run across a multitude of MCUs and hence is preferable to a vendor specific naming.

aurelj:
I still think more generic names would be better, but to be really useful this has to be done at the embedded-hal level. I hope this will be done at some point, but in the meantime I think it is better to merge this PR, which provide a useful feature, rather than leaving it floating around for a long time.

So unless @therealprof  disagrees about this, I will merge it soon.

therealprof:
@aurelj Fully agree with everything you said.

aurelj:
Merged ! Thanks for keeping up with our requests.

PR #69, opened 2019-02-01
Gpio external interrupt

arkap:
I've introduced a new GPIO trait called `ExtiPin` with functions to configure, enable, and disable interrupts from a pin. Pins configured as Input implement this trait.

 - `trigger_on_edge()` sets the necessary bits in the FTSR and/or RTSR to select whether the interrupt is generated on rising, falling, or both edges. There is also an enum (`Edge`) for the three cases.
 - `enable_interrupt()` sets the interrupt enable bit in the EXTI interrupt mask register for the pin and configures the correct GPIO port for that pin in the SYSCFG_EXTICR.
 - `disable_interrupt()` clears the interrupt enable bit in the EXTI interrupt mask register.

I was only able to test this on an stm32f407 but since the register names are the same across all stm32f4 cores I assume this works -- it compiles without name errors for all the devices, anyway.

I welcome any discussion on the interface design be it about the structure or the names of the traits and functions.
For example, it might be convenient to have a function to clear the interrupt pending bit for a given pin.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -80,6 +80,7 @@ pub trait ExtiPin {
     fn trigger_on_edge(&mut self, exti: &mut EXTI, level: Edge);
     fn enable_interrupt(&mut self, exti: &mut EXTI);
     fn disable_interrupt(&mut self, exti: &mut EXTI);
+    fn clear_interrupt_pending_bit(&mut self, exti: &mut EXTI);
 }
 
 macro_rules! gpio {
@@ -557,6 +558,11 @@ macro_rules! gpio {
                     /// Disable external interrupts from this pin
                     fn disable_interrupt(&mut self, exti: &mut EXTI) {
                         exti.imr.modify(|r, w| unsafe { w.bits(r.bits() & !(1 << $i)) });
+                    }
+
+                    /// Clear the interrupt pending bit for this pin
+                    fn clear_interrupt_pending_bit(&mut self, exti: &mut EXTI) {
+                        exti.pr.modify(|r, w| unsafe { w.bits(r.bits() | (1 << $i)) });
                     }
                 }
 


src/gpio.rs
```diff
@@ -502,6 +518,43 @@ macro_rules! gpio {
...
+                            Edge::RISING_FALLING => {
+                                exti.rtsr.write(|w| { w.$tri().set_bit() });
+                                exti.ftsr.write(|w| { w.$tri().set_bit() });
+                            }
+                        }
+                    }
+
+                    /// Enable external interrupts from this pin.
+                    /// Note: This configures EXTI line $i to trigger from this pin
+                    fn enable_interrupt(&mut self, exti: &mut EXTI, syscfg: &mut SYSCFG) {
```
aurelj: I'm feeling like it might be better to not need syscfg as an argument here, and instead do the same as we do for RCC everywhere :
```
let syscfg = unsafe { &(*SYSCFG::ptr()) };
```
Note that I have not really thought about all the implications it may have, so I may be wrong about this.
Also note that the same remark might also apply to the exti argument, but I'm even less sure about it.

src/gpio.rs
```diff
@@ -502,6 +518,43 @@ macro_rules! gpio {
...
+                        syscfg.$exticri.modify(|r, w| unsafe {
+                            let mut exticr = r.bits();
+                            exticr = (exticr & $extigpionr) | $extigpionr;
```
aurelj: Unless I'm missing something, this line looks pretty wrong to me. It is actually equivalent to simply writing `exticr = $extigpionr`.

I guess you would need something like this instead:
```
let offset = 4 * ($extigpionr % 4);
exticr = (exticr & !(0xF << offset)) | ($extigpionr << offset);
```

src/gpio.rs
```diff
@@ -828,58 +881,58 @@ gpio!(GPIOH, gpioh, gpiohen, PH, [
...
 gpio!(GPIOK, gpiok, gpioken, PK, [
-    PK0: (pk0, 0, Input<Floating>),
-    PK1: (pk1, 1, Input<Floating>),
-    PK2: (pk2, 2, Input<Floating>),
-    PK3: (pk3, 3, Input<Floating>),
-    PK4: (pk4, 4, Input<Floating>),
-    PK5: (pk5, 5, Input<Floating>),
-    PK6: (pk6, 6, Input<Floating>),
-    PK7: (pk7, 7, Input<Floating>),
+    PK0: (pk0, 0, Input<Floating>, exticr1, 9, tr0, mr0),
```
aurelj: `$extigpionr` should be 10 here instead of 9.

src/gpio.rs
therealprof: Actually we might have been a bit too lax here in the past. For RMW operations we should always require either a `&mut` on the resource or a `CriticialSection` , potentially we could also offer both options as a convenience.

src/gpio.rs
therealprof: Thanks for catching this. Indeed the code is wrong here.

src/gpio.rs
arkap: You mean a safe function that requires the SYSCFG and an unsafe one which works like @aurelj suggested?

arkap: True. What do you think about introducing another function handling only the `SYSCFG` part of the configuration? This would separate interrupt configuration from enabling and disabling them. The new function then only needs access to `SYSCFG` leaving only `EXTI` for `enable_interrupt()`.

therealprof: @arkap Sounds fantastic.

PR #72, opened 2019-02-06
Gpio exti refactor

arkap:
I've tweaked my recent external interrupt pin implementation to reduce the macro invocation size and cover partially erased pins.
Since the latter know their pin number all the necessary offsets in config registers can be computed.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -172,30 +172,22 @@ macro_rules! gpio {
                     match self.i {
                         0...3 => {
                             syscfg.exticr1.modify(|r, w| unsafe {
-                                let mut exticr = r.bits();
-                                exticr = (exticr & !(0xf << offset)) | ($extigpionr << offset);
-                                w.bits(exticr)
+                                w.bits((r.bits() & !(0xf << offset)) | ($extigpionr << offset))
                             });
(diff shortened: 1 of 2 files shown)


src/gpio.rs
```diff
@@ -165,6 +165,77 @@ macro_rules! gpio {
...
+                                w.bits(exticr)
+                            });
+                        },
+                        12...15 => {
+                            syscfg.exticr4.modify(|r, w| unsafe {
+                                let mut exticr = r.bits();
+                                exticr = (exticr & !(0xf << offset)) | ($extigpionr << offset);
+                                w.bits(exticr)
+                            });
+                        },
```
aurelj: This whole code repetition is slightly ugly, but I see no solution to avoid it.
At least you could drop the exticr intermediate variable to shorten the code to simply:
```
w.bits((r.bits() & !(0xf << offset)) | ($extigpionr << offset))
```

therealprof:
Oh, it would be nice if you could add a CHANGELOG.md entry for your additions.

arkap:
Done, thanks for pointing that out.

therealprof:
Thanks a lot.

PR #73, opened 2019-02-09
Various GPIO improvements

aurelj:


aurelj:
> What would you think about having a release after merging this?

I just realize that we had no release with support for the whole stm32f4 family yet... So yes, indeed it would be great to have a release now. I will send a PR to update the changelog with a few more missing entry and then I think we should release.

PR #81, opened 2019-03-05
Hal ADC supporting one-shot and sequence conversion of regular channels.

cs2dsb:
Hi,

I've taken a bash at doing an ADC implementation.

It's got embedded_hal::adc::Channel implementations for all VREF, temp sensor and VBAT and of course the GPIO pins. Getting a complete list of all the GPIO and which ADC they are hooked to was a bit of a faff but eventually I used a few grep commands to pull the info out of the CubeMX database files so it should be pretty accurate. CubeMX claims that all STM32F4 devices have the same ADC IP block so it should function the same across all devices this crate supports. 

The current device features aren't sufficient to perfectly identify which GPIO are available because some device variants will have some pins missing or added. However, these are in the minority so I don't think it's worth adding another 10+ features for specific device variants. If someone wants to go that route later I've marked the variant specific GPIO with comments so it should be pretty straightforward.

I've implemented one-shot and sequences of regular conversions, DMA config, external triggers and most of the available configuration. I haven't implemented injected or discontinuous conversions or the analog watchdog functionality. 

I've tested one-shot, sequences, DMA and external triggering on an STM32F411 and it works as expected. There are examples of everything except DMA in the documentation for the adc::Adc struct.

I'm going to take a look at an implementation of DMA next but wanted to get this pull request up for feedback sooner rather than later plus the ADC stuff is pretty stand alone.

I ran tools/check.py and clippy so it should at least compile ;)

I'm working on a hardware project using an F4 so I'll have a bunch of other stuff to contribute if all goes well.

Please take a look and let me know if you have any feedback. 

Cheers, 

Daniel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Hal ADC supporting one-shot and sequence conversion of regular channels.
+
 ## [v0.3.0] - 2019-01-14
 
 ### Added
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -1,0 +1,1639 @@
+//! Analog to digital converter configuration.
+//! According to CubeMx, all STM32F4 chips use the same ADC IP so this should be correct for all variants.
+
+#![deny(missing_docs)]
+
+/// This is the address the VREF calibration value is stored in at the factory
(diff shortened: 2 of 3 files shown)


astro:
:+1: Works for me (one-shot mode on STM32F429)

cs2dsb:
I've added it to the changelog and pushed the new commit.

Cheers

PR #82, opened 2019-03-06
Spi interrupt config

arkap:
Extend the SPI API to enable and disable interrupts, and to enable and disable the peripheral alltogether.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -705,6 +705,16 @@ pins! {
         MOSI: [PC1<Alternate<AF5>>]
 }
 
+/// Interrupt events
+pub enum Event {
+    /// New data has been received
+    Rxne ,
+    /// Data can be sent
+    Txe,
+    /// An error occurred
+    Error,
+}
+
 #[derive(Debug)]
 pub struct Spi<SPI, PINS> {
     spi: SPI,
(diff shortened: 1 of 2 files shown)


src/spi.rs
```diff
@@ -779,6 +779,71 @@ macro_rules! hal {
                     Spi { spi, pins }
                 }
 
+                /// Enable transmit register empty interrupt
+                pub fn enable_txe_interrupt(&mut self) {
+                    self.spi.cr2.write(|w| { w.txeie().set_bit() });
```
therealprof: That does not look right, you'll have to use a read-modify-write operation, otherwise you're clearing all other flags.

arkap:
Oh damn, sorry. Lesson learned: Test even the seemingly simplest one-liners.

src/spi.rs
```diff
@@ -779,6 +779,65 @@ macro_rules! hal {
                     Spi { spi, pins }
                 }
 
+                /// Enable transmit register empty interrupt
+                pub fn enable_txe_interrupt(&mut self) {
+                    self.spi.cr2.modify(|r, w| unsafe {w.bits(r.bits() | (1 << 7))});
```
therealprof: This should work, but a more canonical (and saner) version would be:

```suggestion
                    self.spi.cr2.modify(|_, w| { w.txeie().set_bit() });
```

You only need to use the `r` proxy part if you're actually interested in the value e.g. for decision making or because the SVD file is broken so you have to work with the raw value.

src/spi.rs
```diff
@@ -779,6 +779,65 @@ macro_rules! hal {
...
+                /// Enable the peripheral, i.e. set the SPE bit in the SPI control register 1
+                pub fn enable(&mut self) {
+                    self.spi.cr1.modify(|r, w| unsafe {w.bits(r.bits() | (1 << 6))});
+                }
+
+                /// Halt the SPI according to the procedure described in chapter
+                /// 28.3.8 of the STM32F4xx reference manual
+                // TODO: this only applies to master bi-directional transfer mode
+                pub fn wait_until_idle(&mut self) {
+                    while self.spi.sr.read().rxne().bit_is_clear() {}
```
therealprof: The more traditional way here would be to only have one loop, one read and three checks in one go breaking the loop when all conditions are fulfilled.

src/spi.rs
arkap: Oh, that's much prettier. I didn't know this works.

src/spi.rs
arkap: The STM32f4 Reference Manual implies an order to the checks:
manual

I also think another function that allows skipping the `rxne` check might be handy. I see an issue with applications that need to read the received data before they know if the transfer is complete. With the API as it is in the PR, this would go into an infinite loop:

```rust
let data = spi.read();    // clears rxne bit
if transfer_done(data) {
    spi.wait_until_idle(); // rxne cleared at this point -> infinite loop
    spi.disable();
}
```
What do you think?

therealprof: It is weird that there's a fixed order with separate readings here especially since those events could trigger again in the meantime. I didn't quite get the point of `wait_for_idle()`: Ideally we don't have to use it at all unless we want to disable the peripheral, but in that case it really should be done automatically to prevent usage errors.

arkap: Ok, I agree. How about I remove `enable()`, `disable()`, and `wait_until_idle()` from this PR alltogether and we see if we need it in the future?

therealprof: Works for me.

therealprof:
One thing I'd like to bring up for discussion is whether this is the interface we'd like to have. I don't have any particular preference but in various places we have an interface like so: https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/serial.rs and it might make sense to settle on a single interface to simplify moving between chips.

CC @aurelj

arkap:
I'm all for unifying the code and don't see any problem with the `listen`/`unlisten` interface. It does not allow to atomically enable or disable all 3 interrupts, though, but I don't think that is required.

arkap:
There is one more thing: The referenced serial API only allows  listening for `TXE` and `RXNE` eventss and not the various errors. I would prefer to cover the whole range of possible events. Unless there is some reason for this that I don't see.

therealprof:
The ability to generate interrupts for all possible events sounds good to me.

src/spi.rs
```diff
@@ -705,6 +705,16 @@ pins! {
         MOSI: [PC1<Alternate<AF5>>]
 }
 
+/// Interrupt events
+pub enum Event {
+    /// New data has been received
+    Rxne ,
```
therealprof:
```suggestion
    Rxne,
```

Please make sure to run `rustfmt` on your code.

src/spi.rs
```diff
@@ -779,6 +789,48 @@ macro_rules! hal {
...
+
+                /// Return `true` if the RXNE flag is set, i.e. new data has been received
+                /// and can be read from the SPI.
+                pub fn is_rxne(&self) -> bool {
+                    self.spi.sr.read().rxne().bit_is_set()
+                }
+
+                /// Return `true` if the MODF flag is set, i.e. the SPI has experienced a
+                /// Master Mode Fault. (see chapter 28.3.10 of the STM32F4 Reference Manual)
+                pub fn is_modf(&self) -> bool {
```
therealprof: Is `MODF` the only possible error? Would be nice to have some symmetry between enabled events and flags. Other than that and the formatting nit this looks good to me.

arkap: With the SPI implementation as it is (Master mode, two unidirectional lines) this is the only error that can happen, if I understand the reference manual correctly. The other errors only apply to TI mode, slave mode or I2S mode.

therealprof:
Thanks!

PR #86, opened 2019-03-14
rcc: fix rounding of hclk and pclk divisor

### LINKED ISSUE #85: Cannot use USART 1 of STMF407G (DISCOVERY)
I have tried to setup and use the USART 1 for serial communication. If the baud rate is 1200/2400, it works; around 4800 communication becomes corrupted and 19k2 or above just doesn't seem to work anymore.  I have adapted the "serial_echo" example from the stm32f407g-disc create to work on this HAL crate as follows:

```rust
#![no_main]
#![no_std]

extern crate cortex_m_rt;
extern crate panic_semihosting;

use cortex_m_rt::entry;
use hal::block;
use hal::prelude::*;
use hal::serial::{config::Config, Serial};

#[entry]
fn main() -> ! 
    if let Some(p) = hal::stm32::Peripherals::take() {
        let gpioa = p.GPIOA.split();
        let rcc = p.RCC.constrain();
        let clocks = rcc.cfgr.sysclk(48.mhz()).freeze();

        // USART 1 at PA9 (TX) and PA10(RX)
        let tx = gpioa.pa9.into_alternate_af7();
        let rx = gpioa.pa10.into_alternate_af7();

        // Set up USART 1 configured pins and a baudrate of 115200 baud
        let serial = Serial::usart1(
            p.USART1,
            (tx, rx),
            Config::default().baudrate(115_200.bps()),
            clocks,
        )
        .unwrap();

        // Separate out the sender and receiver of the serial port
        let (mut tx, mut rx) = serial.split();

        loop {
            // Read character and echo it back
            let received = block!(rx.read()).unwrap();
            block!(tx.write(received)).ok();
        }
    }

    loop {
        continue;
    }
}
```

But this results in a panic, even though we are only configuring the sysclk:
```
panicked at 'assertion failed: pclk1 <= pclk1_max', /home/paul/.cargo/registry/src/github.com-1ecc6299db9ec823/stm32f4xx-hal-0.3.0/src/rcc.rs:277:9
```

When leaving out the `sysclk(48.mhz())` call, it compiles, but communication seems inpossible.
When setting to 9600 baud and sending 0xAA (10101010), I get back 0xFA or 0xEA.

See also the attached output below of a logic analysis where you can see that the TX part has malformed with respect to what was sent on RX.

Screenshot from 2019-03-08 22-32-11

When changing the code to use USART 2 (pins PA1/PA2) instead, everything works just fine.

aurelj:
fix #85

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -221,7 +221,8 @@ impl CFGR {
 
         assert!(!use_pll || sysclk <= sysclk_max && sysclk >= sysclk_min);
 
-        let (hpre_bits, hpre_div) = match sysclk / self.hclk.unwrap_or(sysclk) {
+        let hclk = self.hclk.unwrap_or(sysclk);
+        let (hpre_bits, hpre_div) = match (sysclk + hclk - 1) / hclk {
             0 => unreachable!(),
             1 => (HPREW::DIV1, 1),
             2 => (HPREW::DIV2, 2),
@@ -265,7 +266,7 @@ impl CFGR {
         let (pclk1_max, pclk2_max) = (50_000_000, 100_000_000);
 
         let pclk1 = self.pclk1.unwrap_or(core::cmp::min(pclk1_max, hclk));
-        let (ppre1_bits, ppre1) = match hclk / pclk1 {
+        let (ppre1_bits, ppre1) = match (hclk + pclk1 - 1) / pclk1 {
             0 => unreachable!(),
             1 => (0b000, 1),
             2 => (0b100, 2),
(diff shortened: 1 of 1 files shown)


src/rcc.rs
```diff
@@ -280,7 +281,7 @@ impl CFGR {
         assert!(pclk1 <= pclk1_max);
 
         let pclk2 = self.pclk2.unwrap_or(core::cmp::min(pclk2_max, hclk));
-        let (ppre2_bits, ppre2) = match hclk / pclk2 {
+        let (ppre2_bits, ppre2) = match (hclk + pclk2 -1) / pclk2 {
```
paulvt: To be very nitpicky, this should probably be:
```suggestion
        let (ppre2_bits, ppre2) = match (hclk + pclk2 - 1) / pclk2 {
```

aurelj: Oh, good catch. Thanks for the review. Fixed.

PR #88, opened 2019-03-16
 Attempt at DMA configuration implementation

cs2dsb:
This adds some (most...?) configuration options for F4 DMA peripherals.

It attempts to enforce valid combinations of DMA peripheral, Stream, Channel, Src, Dest and Direction. The valid combinations have been extracted from CubeMx.

The only combination that I've tested to a reasonable level is the ADC (since that's what I'm using) but from looking at the STM HAL c code I think a lot of the other peripherals I've implemented should work. If they don't the fixes should be minimal (changing the Address trait implementations for the offending peripherals.)

I haven't attempted to implement peripherals my device doesn't have or ones with complex config (like AES, SDIO, DCMI). There are also some peripherals that appear to be missing SVD defines for their DMA registers so there'll probably be some stm32-rs work to make those work. The combinations I haven't implemented are commented out in dma.rs so whoever ends up implementing them shouldn't need to work out the valid Stream/Channel mappings again.

For peripherals with multiple source/dest registers I've created some new types like CCR1, CCR2 which you use to indicate which register DMA should access CCR1(TIM1) -> tim.ccr1, etc.

Upside of this approach is you get compile time checking that your DMA config is possible. Downside is the type signature is a bit gnarly:
```
DmaStream::<DMA2, Stream0<DMA2>, Channel0, ADC1, PeripheralToMemory>::init(...)
```

It's also possible to ignore the type checking and directly use the StreamX structs:
```
Stream0::<DMA2>::clear_interrupts(dma2)
```

I haven't made any attempt to implement a safe API for managing buffers - you just pass in &[u16] or &u32. It's also possible to use custom types if you implement the Address trait - one of my test projects had structs like:
```
#[repr(C)]
pub struct Adc1Samples {
  a: u16,
  b: u16,
  ...
}
```
Which felt like a nice pattern for pulling out each sample from a contiguous lump of memory.

Once some hal traits for DMA emerge it should be straightforward to implement them on the StreamX or DmaStream types.

Hopefully this is a useful starting point :)

### DIFF
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -91,7 +91,7 @@ use crate::{
 };
 use core::{
     marker::{PhantomData, Sized},
-    ops::Deref,
+    ops::{Deref, DerefMut},
 };
 
 #[derive(Debug, Clone, Copy, PartialEq)]
@@ -239,6 +239,7 @@ pub trait Direction: Bits<u8> {}
 #[derive(Debug, Clone, Copy)]
 pub struct PeripheralToMemory;
 impl Bits<u8> for PeripheralToMemory {
+    #[inline]
     fn bits(self) -> u8 {
         0
     }
@@ -248,6 +249,7 @@ impl PeripheralToMemory {
     fn new() -> Self {
         PeripheralToMemory
     }
+    #[inline]
     fn direction() -> DmaDirection {
         DmaDirection::PeripheralToMemory
     }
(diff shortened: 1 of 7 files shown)


src/dma.rs
```diff
@@ -0,0 +1,1999 @@
...
+
+                /// Set the fifo threshold (fcr.fth) of the DMA stream
+                pub fn set_fifo_threshold(dma: &mut $DMA, fifo_threshold: config::FifoThreshold) {
+                    dma.$fcr.modify(|_, w| w.fth().bits(fifo_threshold.bits()));
+                }
+
+                /// Enable/disable the fifo (dmdis) of the DMA stream
+                pub fn set_fifo_enable(dma: &mut $DMA, fifo_enable: bool) {
+                    //Register is actually direct mode disable rather than fifo enable
+                    dma.$fcr.modify(|_, w| w.dmdis().bit(!fifo_enable));
```
MabezDev: Just passing through, I've adapted your PR for a F2 board I am using and its working well, Thanks!

Only one issue I had was in `set_fifo_enable`. This should be just `fifo_enable`, inverting does the opposite.

thalesfragoso:
Hey @cs2dsb , I have some ideas for modifications and additions to make a safe interface for this, I will try to start working on this soon, can you give me access to your branch ?

cs2dsb:
> Hey @cs2dsb , I have some ideas for modifications and additions to make a safe interface for this, I will try to start working on this soon, can you give me access to your branch ?

I'm not particularly familiar with how the access stuff works in github but I've invited you as a collaborator to my fork so hopefully that'll let you do what you want there. If not, I'd be fine with you cherry picking/merging the commits I made into your fork then I'll delete this PR. I'm not currently working with the F4 so I'm unlikely to contribute to it for a while.

thalesfragoso:
I will be closing this here, since most of the work was merged in #186, I would gladly review a PR to bring in your ADC additions.

Thanks for your work.

PR #89, opened 2019-03-18
implement device electronic signature

astro:
This data is not a peripheral (hence not in SVDs) but in flash memory.

Example use case: generating unique stable device addresses for network communications.

Should exist across the whole f4xx line.

### DIFF
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -68,3 +68,42 @@ impl FlashSize {
         usize::from(self.kilo_bytes()) * 1024
     }
 }
+
+/// ADC VREF calibration value is stored in at the factory
+#[derive(Debug)]
+#[repr(C)]
+pub struct VrefCal(u16);
+define_ptr_type!(VrefCal, 0x1FFF_7A2A);
+
+impl VrefCal {
+    /// Read calibration value
+    pub fn read(&self) -> u16 {
+        self.0
+    }
+}
+
+/// A temperature reading taken at 30°C stored at the factory
+#[derive(Debug)]
+#[repr(C)]
+pub struct VtempCal30(u16);
+define_ptr_type!(VtempCal30, 0x1FFF_7A2C);
+
+impl VtempCal30 {
+    /// Read calibration value
+    pub fn read(&self) -> u16 {
+        self.0
+    }
+}
+
+/// A temperature reading taken at 110°C stored at the factory
+#[derive(Debug)]
+#[repr(C)]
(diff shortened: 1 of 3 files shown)


src/signature.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /// Wafer number
+    pub fn waf_num(&self) -> u8 {
+        self.waf_lot[0].into()
+    }
+
+    /// Lot number
+    pub fn lot_num(&self) -> &str {
+        from_utf8(&self.waf_lot[1..]).unwrap()
```
aurelj: It might be better to use from_utf8_unchecked() rather than using unwrap() ? And it is supposed to be more lightweight as well, which might be important on a microcontroller.

aurelj:
Also, I think VREFCAL, VTEMPCAL30 and VTEMPCAL110 which are currently defined in adc.rs should be part of this signature.rs module as they are also stored in the electronic signature flash block.

src/signature.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// Wafer number
+    pub fn waf_num(&self) -> u8 {
+        self.waf_lot[0].into()
+    }
+
+    /// Lot number
+    pub unsafe fn lot_num(&self) -> &str {
+        from_utf8_unchecked(&self.waf_lot[1..])
+    }
```
aurelj: Any issue with putting the from_utf8_unchecked() inside an unsafe block rather than making the function unsafe ?

astro: Is ASCII from flash potentially unsafe (corrupted) in the real world?

astro:
> Also, I think VREFCAL, VTEMPCAL30 and VTEMPCAL110 which are currently defined in adc.rs should be part of this signature.rs module as they are also stored in the electronic signature flash block.

Added! If you like it, we should update `adc.rs` to make use of them…

src/signature.rs
aurelj: I have no idea. Ideally we would check with as much different STM32 as possible, but I guess for now we should just trust the ST Reference Manual which specify it is ASCII. If we encounter some example with non-ASCII in the future, we will think about a solution then.

aurelj:
> Added! If you like it, we should update `adc.rs` to make use of them…

Nice ! Yes, I like it overall.
And yes, `adc.rs` should be updated, but that's a separate issue, and should probably be done in another PR after this one is merged.

PR #91, opened 2019-04-02
Send NACK on I2C read

timothyhollabaugh:
When writing a driver for the [VL6180x rangefinder](https://www.st.com/en/imaging-and-photonics-solutions/vl6180x.html), I noticed that my stm32f405rg was not sending a NACK at the end of a read sequence. This caused the VL6180x to send an extra byte before the clock stopped, but more importantly, it left the SDA line pulled low for the next I2C operation. This caused the I2C peripheral to think the line was busy, thus not allowing any bytes to be sent.

At first, I fixed this by resetting the I2C peripheral and manually setting both the SDA and SCL lines to a high, open drain configuration after a read. However, I noticed that the stm32f405 reference guide specifies a rather specific sequence to properly send the NACK at the end of a sequence of reading bytes. This can be found on page 850 of the reference guide. This involves clearing the ACK bit during the right byte being read, which depends on the total number of bytes being read. Currently, the I2C read method does not even attempt to send a NACK, so I implemented what the datasheet specifies in the read method. It have been tested with the stm32f405 and a VL6180x, reading 16, 8, 4, 3, 2, and 1 byte(s) of data repeatedly.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -838,13 +838,13 @@ where
         if buffer.len() <= 2 {
 
             // Clear the ack flag and set the pos flag to the nack gets sent
-            // after the next recieved byte
+            // after the next received byte
             self.i2c.cr1.modify(|_, w| w.ack().clear_bit().pos().set_bit());
 
             // Clear addr by reading SR2
             self.i2c.sr2.read();
 
-            // Wait untill both bytes are recived. Byte 1 will be in dr,
+            // Wait until both bytes are received. Byte 1 will be in dr,
             // and byte 2 will be in the shift register
             while {
                 let sr1 = self.i2c.sr1.read();
@@ -895,7 +895,7 @@ where
             // Set the stop bit to finish the transaction
             self.i2c.cr1.modify(|_, w| w.stop().set_bit());
 
-            // Recieve the last 2 bytes
+            // Receive the last 2 bytes
             last[1] = self.recv_byte()?;
             last[2] = self.recv_byte()?;
         }
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -835,17 +835,70 @@ where
             sr1.addr().bit_is_clear()
         } {}
 
-        // Clear condition by reading SR2
-        self.i2c.sr2.read();
+        if buffer.len() <= 2 {
 
-        // Receive bytes into buffer
-        for c in buffer {
-            *c = self.recv_byte()?;
+            // Clear the ack flag and set the pos flag to the nack gets sent
+            // after the next recieved byte
+            self.i2c.cr1.modify(|_, w| w.ack().clear_bit().pos().set_bit());
+
+            // Clear addr by reading SR2
+            self.i2c.sr2.read();
+
+            // Wait untill both bytes are recived. Byte 1 will be in dr,
+            // and byte 2 will be in the shift register
+            while {
+                let sr1 = self.i2c.sr1.read();
+                sr1.btf().bit_is_clear()
+            } {}
+
+            // Set the stop bit to finish the transaction
+            self.i2c.cr1.modify(|_, w| w.stop().set_bit());
+
+            // Move both bytes into the buffer
+            for c in buffer {
+                *c = self.recv_byte()?;
+            }
+        } else {
+
+            let n = buffer.len();
+
+            let (first, last) = buffer.split_at_mut(n-3);
+
+            // Clear addr by reading SR2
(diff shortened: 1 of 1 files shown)


src/i2c.rs
```diff
@@ -835,17 +835,70 @@ where
...
-            *c = self.recv_byte()?;
-        }
+            // Clear the ack flag and set the pos flag to the nack gets sent
+            // after the next recieved byte
+            self.i2c.cr1.modify(|_, w| w.ack().clear_bit().pos().set_bit());
+
+            // Clear addr by reading SR2
+            self.i2c.sr2.read();
+
+            // Wait untill both bytes are recived. Byte 1 will be in dr,
```
therealprof:
```suggestion
            // Wait until both bytes are received. Byte 1 will be in dr,
```

src/i2c.rs
```diff
@@ -835,17 +835,70 @@ where
...
+            // and the n byte is in the shift register
+            while {
+                let sr1 = self.i2c.sr1.read();
+                sr1.btf().bit_is_clear()
+            } {}
+
+            // Set the stop bit to finish the transaction
+            self.i2c.cr1.modify(|_, w| w.stop().set_bit());
+
+            // Recieve the last 2 bytes
```
therealprof:
```suggestion
            // Receive the last 2 bytes
```

src/i2c.rs
```diff
@@ -835,17 +835,70 @@ where
...
-        // Clear condition by reading SR2
-        self.i2c.sr2.read();
+        if buffer.len() <= 2 {
 
-        // Receive bytes into buffer
-        for c in buffer {
-            *c = self.recv_byte()?;
-        }
+            // Clear the ack flag and set the pos flag to the nack gets sent
+            // after the next recieved byte
```
therealprof:
```suggestion
            // after the next received byte
```

src/i2c.rs
```diff
@@ -835,17 +835,70 @@ where
...
+            self.i2c.cr1.modify(|_, w| w.ack().clear_bit().pos().set_bit());
```
aurelj: According to the reference manual:
```
The POS bit must be used only in 2-byte reception configuration in master mode
```
Here you are setting it even when receiving only one byte.

Also this POS trick is a kind of optimization for receiver that always receive exactly 2 bytes, but it should be possible to receive 2 bytes without using the POS trick, and thus, I think we don't need to have special case for len <= 2.

aurelj:
If my reading of the Reference manual is correct, this much simpler patch (totally untested) should also work :
```
diff --git a/src/i2c.rs b/src/i2c.rs
index 44bc5ba..9254cb4 100644
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -838,13 +838,18 @@ where
         // Clear condition by reading SR2
         self.i2c.sr2.read();
 
+        let (last, buffer) = buffer.split_last_mut().unwrap();
+
         // Receive bytes into buffer
         for c in buffer {
             *c = self.recv_byte()?;
         }
 
-        // Send STOP condition
-        self.i2c.cr1.modify(|_, w| w.stop().set_bit());
+        // Prepare to send NACK then STOP after next byte
+        self.i2c.cr1.modify(|_, w| w.ack().clear_bit().stop().set_bit());
+
+        // Receive last byte
+        *last = self.recv_byte()?;
 
         // Fallthrough is success
         Ok(())
```

Also we should probably check that `buffer.len() > 0` to ensure that the unwrap() can't fail.

timothyhollabaugh:
I was originally going to do something like that, but with `if let Some((last, buffer)) = ...` to catch the empty buffer case, but I wasn't sure what the POS bit did and figured I should probably follow the datasheet exactly. If I get the chance over the next few weeks, I will test out your simplified version and see how it works.

pferreir:
I was having the same issue and tested both patches. Both work OK, but @aurelj's seems to reflect the expected behaviour, with one byte exactly being read.

My code:

```rust
    pub fn get_chip_id_rev(&mut self) -> Result<(u8, u8), <I2C as hal::blocking::i2c::WriteRead>::Error> {
        let mut buffer: [u8; 1] = [0];
        self.i2c.write_read(self.i2c_addr, &[CS43L22Register::ID as u8], &mut buffer)?;
        let d = buffer[0];
        Ok((d >> 3, d & 0x7))
    }
```
@timothyhollabaugh's version:
image


@aurelj's version:

image

timothyhollabaugh:
Since this was fixed in #92, I will close this one.

PR #92, opened 2019-04-23
I2C: properly terminate read with a NACK then a STOP

aurelj:
This PR is an improved alternative to #91, which contains all the details about this change.


aurelj:
> Please add this fix to CHANGELOG.md

Oh, of course, I keep forgetting about the changelog.
Fixed.

PR #95, opened 2019-04-25
spi: use generic impl over spi1::RegisterBlock to avoid the need for macro

aurelj:
With the updates in stm32f4 v0.7.0, we can now apply to SPI the same treatment that was applied to I2C some times ago:
Make use of a generic implementation over spi1::RegisterBlock instead of generating several implementations with a big macro.


aurelj:
Damn ! If only I wasn't the one who initially wrote the CHANGELOG.md file, I could try to find some excuses...
Anyway, patch updated.

PR #97, opened 2019-04-28
added QEI mode support for TIM1, TIM2, TIM3, TIM4, TIM5 and TIM8

ilya-epifanov:


ilya-epifanov:
an Impl day at Oxidize well spent!

src/qei.rs
```diff
@@ -0,0 +1,445 @@
...
+                            .cc1p()
+                            .clear_bit()
+                            .cc2e()
+                            .set_bit()
+                            .cc2p()
+                            .clear_bit()
+                    });
+
+                    // configure as quadrature encoder
+                    #[allow(unused_unsafe)]
```
therealprof: Huh? Unused unsafe?

ilya-epifanov: It's not the first time I hit this bug in rustc.

Without `unsafe {}`:
```
error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
   --> src/qei.rs:809:40
    |
809 |                       tim.smcr.write(|w| w.sms().bits(3));
    |                                          ^^^^^^^^^^^^^^^ call to unsafe function
...
862 | / hal! {
863 | |     TIM1: (tim1, tim1en, tim1rst, apb2enr, apb2rstr, u16),
864 | |     TIM5: (tim5, tim5en, tim5rst, apb1enr, apb1rstr, u32),
865 | | }
    | |_- in this macro invocation
    |
    = note: consult the function's documentation for information on how to avoid undefined behavior
```

aurelj: It most likely is unsafe on one variant of stm32f4 and safe on another, or unsafe for TIM1 and safe for TIM5 for example...

ilya-epifanov: Yep, that's exactly the case. I've added a comment.

ilya-epifanov:
@therealprof @aurelj done

aurelj:
Thanks a lot !

PR #103, opened 2019-07-18
gpio: update to digital::v2 traits from the now deprecated v1 traits

aurelj:


therealprof:
Not quite what I had in mind but sure, let's break the world. 😅

aurelj:
Not sure what you had in mind ?

Looking at other repo, it seems it was done in the exact same way:
https://github.com/stm32-rs/stm32f3xx-hal/commit/5985970194334626a06e5d28303c75ad884c1505
https://github.com/stm32-rs/stm32l0xx-hal/commit/d27f6823408197bec39c494f3a6b586ca962c11a

Yes, this breaks the world, and will require a version bump at next release. This has to be done at some point anyway... And digital::v1_compat can be used in projects that want to update their dependencies with minimal changes.

PR #112, opened 2019-10-19
Add method for clearing TIMx interrupt flags

FLamparski:
When a TIM interrupt fires, it needs to be cleared, otherwise it will immediately retrigger. This patch adds a HAL method to do so, removing the need for unsafe low-level access.

Example:

```rust
/// This variable holds a reference to TIM2, which will be accessed in the ISR in order
/// to clear the interrupt flag.
static TIMER_TIM2: Mutex<RefCell<Option<Timer<stm32::TIM2>>>> = Mutex::new(RefCell::new(None));

fn main () -> ! {
    // ... setup skipped ...
    let mut timer = Timer::tim2(dp.TIM2, 1.khz(), clocks);
    timer.listen(Event::TimeOut);
    // Store the timer reference in the global variable
    free(|cs| *TIMER_TIM2.borrow(cs).borrow_mut() = Some(timer));
}

#[interrupt]
fn TIM2() {
    free(|cs| {
        // Access the global timer reference and clear the timer interrupt
        if let Some(ref mut tim2) = TIMER_TIM2.borrow(cs).borrow_mut().deref_mut() {
            tim2.clear_interrupt(Event::TimeOut);
        }
    })
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Implemented `Qei` trait
+- Implemented `clear_interrupt()` method for TIM timers
 
 ## [v0.5.0] - 2019-04-27
 
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -177,11 +177,27 @@ macro_rules! hal {
                 }
 
                 /// Starts listening for an `event`
+                ///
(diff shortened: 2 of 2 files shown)


FLamparski:
Thanks for merging. When do you think this will be released?

therealprof:
I can do it right now.

FLamparski:
That would be lovely!

therealprof:
https://crates.io/crates/stm32f4xx-hal

PR #113, opened 2019-10-20
Add some examples

FLamparski:
One thing that's bothered me about this crate was the lack of examples - so here are some.

* **delay-blinky** is from https://github.com/FLamparski/stm32f446-blinky
* **ssd1306-image** is from https://github.com/FLamparski/stm32f446-oled-test
* **stopwatch-with-ssd1306-and-interrupts** is from https://github.com/FLamparski/rusted-brains-stopwatch

They were all developed on the Nucleo F446RE board.


therealprof:
Great stuff, thanks. Just for reference, there're more crates with examples, like https://github.com/stm32-rs/stm32f407g-disc and https://github.com/stm32-rs/stm32f429i-disc but it's certainly good to have some directly in the HAL impl crate.

FLamparski:
Note that while the examples here pass _compilation_, the two examples with ssd1306 won't link because they're too big. Should I just bump the defaults in memory.x or do you want to keep that file as conservative as possible?

therealprof:
You mean it doesn't link in debug mode? That's fine. I would assume that 64kB should be enough for release mode.

PR #114, opened 2019-10-22
Fix clear_interrupt_pending_bit()

Windfisch:
According to RM0090 12.3.6, a "bit is cleared by programming it to 1".

I have compiled, but not tested this on stm32f4xx. However, I found this issue while porting the ExtiPin trait to stm32f1xx. I have tested this change on my stm32f1xx-hal working copy.

See also https://github.com/stm32-rs/stm32h7xx-hal/pull/41


therealprof:
Well, I missed it, but the current code will clear all set bits but here's the kicker, the one used in the PR is not quite right as well since`modify()` does a read-modify-write operation while what you want is to do in an explicit and unconditional `write()`.

Windfisch:
> Please use `write()` instead of `modify()`.

fixed; also, I removed the &mut EXTI reference from the method signature, because it is unneccessary.

Issue #99, opened 2019-05-21
The I2c driver does not check/set open_drain


korken89:
Hi,

I was having issues with the I2c and it turned out to be that I had not done `set_open_drain()` on the pins.

There should be a check, or that the I2c should set this - as missing this does not produce a compile error and it can in worst case burn a slave when the master holds the bus high and the slave wants to ACK the address.
As it is now should be considered a design error as it can cause irreparable damage to a system without (compile-time/run-time) warning.

BR Emil

therealprof:
There're many ways to destroy your hardware by improper use, even by use of software and there's nothing one can do about it. In this particular case we could probably address it and I wouldn't object to having a compile time error if someone doesn't put the pin in the right state for I2C operation.

Any takers to create a PR? 😉

VersBinarii:
I would love to help with this but i would be grateful for some guidance as i'm not exactly sure how to implement this as a compile-time check.

therealprof:
@VersBinarii There're a few different ways to address this. I haven't looked into the details so I'm just citing from the top of my head.

- First idea: We're already using type states to make sure that when we hand a GPIO pin over to the I2C initialisation it is in the correct mode. One of the possible ways to address this would be to extend the type states to not only include the alternate function assignment but also the pin drive mode.
- Second idea: We could simply initialise the pin correctly in the I2C initialisation, since we have full control over the pin at that point and other drive modes do make make too much sense for I2C operation.

So the first idea would be the option where the compiler checks the state for us, disadvantage: More fiddly code to handle for the maintainers of the crate. The second idea is where we simply do the right thing(tm) behind the users back, disadvantage: Might require more input parameters to implement safely.

VersBinarii:
@therealprof Yes, i though about the second approach as well because it seemed to me that the first approach would become a bit messy. I also didn't see this being done in any other HALs i saw.

I just thought that the compile time check would be more Rusty and therefore preferable.

bfrog:
It seems like if your are setting the pin for an alternative function (i2c) the crate should appropriately setup the pin drive mode for you for that alternative function with a sensible default. I ran into this myself, and its less than obvious from the API that it's up to me to set that up appropriately.

Saying that I think the second option would be my preferred take on this.

therealprof:
@bfrog Mind trying a PR to implement this?

bfrog:
@therealprof I'll try to find some time tomorrow to do this

therealprof:
[...]

PR #120, opened 2019-12-20
Fix examples, test them in CI

### LINKED ISSUE #116: Adjust CI to also compile examples
As witnessed by #115 we're not catching API changes since we're not building our examples in CI.

jamwaffles:
Tested on an STM32F411 Nucleo dev board.

Solution found in #99 to get the examples to actually show something on the display. The Travis command to test the examples should obviously be folded into the Python check script - can someone help me do that? I have zero Python knowledge.

Closes #116 

Cc jamwaffles/ssd1306#100

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,6 @@ matrix:
 install:
   - rustup target add thumbv7em-none-eabihf
 
-script: tools/check.py $COMMAND
+script:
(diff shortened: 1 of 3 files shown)


therealprof:
Just noticed that some squashing would have been a good idea.  🤷🏻‍♂️

PR #121, opened 2020-01-13
Add wrapper around SYSCFG

jkristell:
Move the enabling off the syscfg clock to the new Syscfg wrapper.

With inspiration from stm32l0xx-hal crate. Tested on stm32f401 with external interrupt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Added examples in the examples folder.
+- [breaking-change] Added a wrapper around SYSCFG.
 
 ## [v0.6.0] - 2019-10-19
 
--- a/examples/stopwatch-with-ssd1306-and-interrupts.rs
+++ b/examples/stopwatch-with-ssd1306-and-interrupts.rs
@@ -14,6 +14,7 @@ use crate::hal::{
     prelude::*,
     rcc::{Clocks, Rcc},
     stm32,
+    syscfg,
     timer::{Event, Timer},
 };
 use arrayvec::ArrayString;
@@ -44,8 +45,6 @@ enum StopwatchState {
 #[entry]
 fn main() -> ! {
     if let (Some(mut dp), Some(cp)) = (stm32::Peripherals::take(), cortex_m::Peripherals::take()) {
-        dp.RCC.apb2enr.write(|w| w.syscfgen().enabled());
-
         let rcc = dp.RCC.constrain();
         let clocks = setup_clocks(rcc);
         let gpiob = dp.GPIOB.split();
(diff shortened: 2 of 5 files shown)


therealprof:
Can you explain the motivation behind this? Seems like if we add such an abstraction we should fully use it instead of providing crate access to the raw register after construction (especially since the constructor does something unrelated in an unsafe way with the RCC)?

jkristell:
I think users of the hal and of the gpio external interrupt functionality of the hal expect it to work without figuring out how to enable the clock to the syscfg in this clumsy way that is needed today (before the rcc is constrained()).

I did a brief investigation on how the other stm32 hals handles this, stm32f1xx-hal has a nice but totally different architecture (compared to this hal) with the Enable Trait.

The stm32l0xx-hal has a similar architecture compared to this crate, and they do protect the access to the RCC in a wrapper, but in this crate there is none of that, just lots of `let rcc = unsafe { &(*RCC::ptr()) };` everywhere. so I went the same route.

It would probably be a good idea to take inspiration from stm32l0xx-hal and wrap the RCC, and I could give it a go. But I don’t think this addition to the hal makes it that much worse :).

therealprof:
I guess at some point we will need to offer a proper interface to enable/disable various blocks in the chips. I haven't yet looked into whether it makes sense to combine SYSCFG and RCC into a single interface but it kind of sounds like it.

jkristell:
The exti configuration, and i/o compensation could probably be combined with the gpio module.

... But then it handles a couple of other tasks as well. 

From the reference manual: 

"The system configuration controller is mainly used to remap the memory accessible in the
code area, select the Ethernet PHY interface and manage the external interrupt line
connection to the GPIOs."

So the reason for enabling the SYSCFG could be any of those i guess. So maybe it better to keep it as a separate module.

jkristell:
Created a branch that changes the subsystems that today dereferences the RCC to instead take a `&mut Rcc` as an argument to their initialization functions.

<https://github.com/jkristell/stm32f4xx-hal/tree/rcc>

Compile tested. Get rids of the unsafes, on the expense of needing to supply Rcc to all the initialization functions. What do you think?

therealprof:
@jkristell Looks good to me, although I'm not too hot on those `rcc.rb` shenanigans.

@aurelj What do you think?

aurelj:
I don't like requiring to supply Rcc to every initialization function. This remove some abstraction from end user, makes the API harder to use and uglier, and gains us no safety at all.

And I think the fact that make_interrupt_source() requires a SYSCFG parameter is an issue as well, for exactly the same reasons. I actually have a patch to remove this in a WIP branch. [Here is just this patch.](https://github.com/aurelj/stm32f4xx-hal/commit/e9b9822) 

Regarding the initial issue, I think we should just add something like this directly in make_interrupt_source():
```
let rcc = unsafe {&*stm32::RCC::ptr()};
rcc.apb2enr.write(|w| w.syscfgen().enabled());
```

The whole point of this crate is abstracting hardware from end user, so exposing the fact that gpio interrupt are actually setup by touching some SYSCFG and RCC registers is kind of defeating the purpose of the crate.

jkristell:
The motivation for this merge request was to make the hal more convenient to use. So enabling the SYSCFG clock in `make_interrupt_source()` is fine by me.

But on the other hand - I can see the need for the exclusive access to the Rcc. If for example the hal gets functionality for enabling/disabling clocks for the subsystems at "runtime". Modifying the rcc from different _threads_, without locking, could lead to race conditions - right?

But right now the Rcc is only modified during the initialization of the peripherals, so I don't think this is a issue.

therealprof:
I also think that we sooner or later need to expose access to the `RCC` in some way or form to allow third party drivers to access low-level functioniality (USB and Ethernet are two prime examples for needing a lot of configuration) or implement sleep modes. The only safe way of doing this is having a handle which can moved around.

If you want to wait until this need actually arises that is fine with me.

jkristell:
I think @aurelj approach is the right one for now. Closing this pull request.

PR #124, opened 2020-02-12
Implement bit-banding

thalesfragoso:
I started a PWM implementation for this HAL and since different channels share the same register we need to make sure that only a single channel modifies the register at a time if we want to split them. So I decided to make this PR to pave the way. Also It can be a solution to the problem of the non-atomically modifying of the `RCC` registers during peripheral initialization.

I would appreciate some opinions about this approach. 
CC @therealprof 

PS: Inspired by the implementation in the F1 HAL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added examples in the examples folder.
 - Added USB driver.
 - PLL48Clk configuration.
+- Added bit-banding implementation.
 
 ## [v0.6.0] - 2019-10-19
 
--- a/src/bb.rs
+++ b/src/bb.rs
@@ -1,0 +1,48 @@
+//! Bit banding
+//!
+//! Support for the manipulation of peripheral registers through bit-banding.
+//! Not all peripherals are mapped to the bit-banding alias region, the peripheral bit-band region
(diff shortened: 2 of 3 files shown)


therealprof:
Huh? What makes you think the F4 supports bit-banding? The only MCU family known to me at the moment which supports bit-banding is F1.

thalesfragoso:
Well, all the manuals listed here
https://github.com/stm32-rs/stm32-rs/blob/master/stm32f4/README.md say that
the chips support bit-banding on peripherals and SRAM and I checked PM0214
to find the addresses. Am I really missing something here ?

Em qua, 12 de fev de 2020 06:21, Daniel Egger <notifications@github.com>
escreveu:

thalesfragoso:
Well, all the manuals listed here https://github.com/stm32-rs/stm32-rs/blob/master/stm32f4/README.md say that the chips support bit-banding on peripherals and SRAM and I checked PM0214 to find the addresses. Am I really missing something here ?

therealprof:
Woah, this is extremely weird. I didn't see it in the memory map of the RM so I thought it did not exist, also I was searching for the wrong keyword so that showed up empty either.

TIL F4 does also support bit-banding.

Sure, your approach should work just fine, then.

thalesfragoso:
Added some docs, marked the functions as `unsafe` and rebased.

PR #126, opened 2020-02-14
Add USB driver

Disasm:


Disasm:
Note that RCC can't setup clocks for USB at the moment, I hope we will see that part in a while. I'm going to add an example as soon as clock part will be ready.

therealprof:
Nice!

Re the clock parts: does that mean you're working on it or are you hoping for someone else to do it?

Disasm:
@therealprof I hope that @TeXitoi will be able to help with clocks, or I will continue working on this on the next week.

Disasm:
The example: https://github.com/Disasm/stm32f4xx-hal/blob/usb-driver-example/examples/usb_serial.rs

TeXitoi:
Yep, I'll try to have a look this week-end

Disasm:
Done!

TeXitoi:
So, I've look at the stm32f401 RM and rcc.rs, and there is something that I don't understand in the actual code.

I can't find any link between https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/rcc.rs#L91-L92 and the RM. This `sysclk_div` is lightly used after, but I don't understand there lines.

Maybe @aurelj can give inputs, as it seems that he is the one that have the most touched to this code?

Then, I propose the following interface:
 - `CFGR::require_pll48clk()` that ask for a ppl48 clock as precise as possible. If used
   - the PLL48 will be <= 48MHz, and the closest possible. 
   - PLL will be running.
 - `Clocks::pll48clk(&self) -> Option<Hertz>` that returns the frequence of the pll48clk line
 - if the user doesn't ask for pll48clk, it will be set <= 48MHz, and the closest possible after choosing the PLL config for optimal SysClk. It will be activated only if the pll is needed (no direct HSE or HSI for SysClk.

So, any feedback before I dive into implementing this?

Thanks.

TeXitoi:
OK, `sysclk_div` is pllp in usable form.

therealprof:
@TeXitoi Some of the code was ported from other families which have a much simpler clock tree. A lot of the shortcomings have been addressed but it's very likely there's a more need for fixing, especially for the higher up models in that line.

Expect the RM (or the STM32CubeMX clock generator) to be a lot closer to reality than the crate. 😅

PR #127, opened 2020-02-15
Require pll48clk

TeXitoi:
Tested here, and ported my keyboard using this. Working as expected.

https://github.com/TeXitoi/keyberon-f4/commit/3375e744b7fb6cd476455a52bd423f4f08184d41

So, writing this with this code.

Some hand testing:

```
            .use_hse(25_000.khz())
            .sysclk(84.mhz())
            .require_pll48clk() // same without this

Clocks {
    hclk: Hertz(
        84000000,
    ),
    pclk1: Hertz(
        42000000,
    ),
    pclk2: Hertz(
        84000000,
    ),
    ppre1: 2,
    ppre2: 1,
    sysclk: Hertz(
        84000000,
    ),
    pll48clk: Some(
        Hertz(
            48000000,
        ),
    ),
}

            .use_hse(25_000.khz())
            .sysclk(83.mhz())
            .require_pll48clk()

clocks = Clocks {
    hclk: Hertz(
        84000000,
    ),
    pclk1: Hertz(
        42000000,
    ),
    pclk2: Hertz(
        84000000,
    ),
    ppre1: 2,
    ppre2: 1,
    sysclk: Hertz(
        84000000,
    ),
    pll48clk: Some(
        Hertz(
            48000000,
        ),
    ),
}


            .use_hse(25_000.khz())
            .sysclk(83.mhz())
            //.require_pll48clk()

clocks = Clocks {
    hclk: Hertz(
        83000000,
    ),
    pclk1: Hertz(
        41500000,
    ),
    pclk2: Hertz(
        83000000,
    ),
    ppre1: 2,
    ppre2: 1,
    sysclk: Hertz(
        83000000,
    ),
    pll48clk: Some(
        Hertz(
            47428571,
        ),
    ),
}


            .use_hse(25_001.khz())
            .sysclk(83.mhz())
            .require_pll48clk()

clocks = Clocks {
    hclk: Hertz(
        83857494,
    ),
    pclk1: Hertz(
        41928747,
    ),
    pclk2: Hertz(
        83857494,
    ),
    ppre1: 2,
    ppre2: 1,
    sysclk: Hertz(
        83857494,
    ),
    pll48clk: Some(
        Hertz(
            47918568,
        ),
    ),
}


            .use_hse(25_001.khz())
            .sysclk(83.mhz())
            //.require_pll48clk()

clocks = Clocks {
    hclk: Hertz(
        82989404,
    ),
    pclk1: Hertz(
        41494702,
    ),
    pclk2: Hertz(
        82989404,
    ),
    ppre1: 2,
    ppre2: 1,
    sysclk: Hertz(
        82989404,
    ),
    pll48clk: Some(
        Hertz(
            47422516,
        ),
    ),
}
```

But I can't make the usb_serial example working, it hangs on my dev board in cortex-m-rt in `pre_init`. I don't have any idea why. Tried several things without any success. @Disasm maybe you have an idea?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ install:
 
 script:
   - tools/check.py $COMMAND
-  - cargo build --examples --features=stm32f411,rt --release
+  - cargo build --examples --features=stm32f401,rt,usb_fs --release
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -18,6 +18,7 @@ impl RccExt for RCC {
                 pclk1: None,
                 pclk2: None,
                 sysclk: None,
+                pll48clk: false,
             },
         }
     }
@@ -36,6 +37,7 @@ pub struct CFGR {
     pclk1: Option<u32>,
     pclk2: Option<u32>,
     sysclk: Option<u32>,
+    pll48clk: bool,
 }
 
 impl CFGR {
@@ -81,12 +83,18 @@ impl CFGR {
         self
     }
 
-    fn pll_setup(&self) -> (bool, u32)
-    {
-        let rcc = unsafe { &*RCC::ptr() };
+    pub fn require_pll48clk(mut self) -> Self {
+        self.pll48clk = true;
+        self
+    }
 
+    fn pll_setup(&self) -> (bool, bool, u32, Option<Hertz>) {
         let pllsrcclk = self.hse.unwrap_or(HSI);
         let sysclk = self.sysclk.unwrap_or(pllsrcclk);
+        let sysclk_on_pll = sysclk != pllsrcclk;
+        if !sysclk_on_pll && !self.pll48clk {
+            return (false, false, sysclk, None);
+        }
 
         // Sysclk output divisor must be one of 2, 4, 6 or 8
         let sysclk_div = core::cmp::min(8, (432_000_000 / sysclk) & !1);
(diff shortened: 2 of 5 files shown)


Disasm:
@TeXitoi Sorry for the late feedback, but I wonder what is the reason behind this getter for the PLL48 clock frequency? This frequency seems to be comparable with the only frequency, why not to make just a boolean field?

Re usb-serial example: have you tried setting HSE to 25 MHz instead of 8?

TeXitoi:
Because with certain crystal, you might not have the exact frequency, and you can check if the deviation is acceptable. USB clock should be precise with a 0.25% precision, thus a small deviation should be OK.

I just tried the version committed, bit it hang strangely before main. I don't see the dbg on the openocd console. That's may only be some miss configuration somewhere.

Disasm:
Well, usb_serial example works for me (f401-black-pill board) with this patch:
```diff
diff --git a/examples/usb_serial.rs b/examples/usb_serial.rs
index 97b49cb..d91245f 100644
--- a/examples/usb_serial.rs
+++ b/examples/usb_serial.rs
@@ -21,7 +21,7 @@ fn main() -> ! {
 
     let clocks = rcc
         .cfgr
-        .use_hse(8.mhz())
+        .use_hse(25.mhz())
         .sysclk(83.mhz())
         .pclk1(24.mhz())
         .require_pll48clk()
```

Without this patch MCU hangs and even openocd can't connect, so I have to enter the bootloader first and then flash.

Disasm:
> Because with certain crystal, you might not have the exact frequency, and you can check if the deviation is acceptable. USB clock should be precise with a 0.25% precision, thus a small deviation should be OK.

If this requirement is documented, maybe we should check this instead of transferring the problem to user?

.travis.yml
```diff
@@ -2,7 +2,6 @@ language: rust
 rust:
   - stable
   - nightly
-cache: cargo
```
therealprof: Why remove the cargo cache?

TeXitoi:
OK, same problem, so it was just that.

.travis.yml
TeXitoi: 1:07 with, 37min without

TeXitoi:
I can add an assert @Disasm if you want. A method `check_pll48clk` on Clocks? And an assert on it if you require pll48?

Disasm:
@TeXitoi maybe `pll48clk_valid() -> bool` that does this bound check

TeXitoi:
I tried to cut the work in independent commit so I can remove any one if you want (came go cache, rust fmt...)

TeXitoi:
OK, I'll add the method and assert when I'll have the time, hopefully this evening.

TeXitoi:
example corrected and `Clocks::is_pll48clk_valid` + assert

.travis.yml
therealprof: That is unexpected and weird. I don't see any reason why it would be faster without.

We can certainly experiment with that but please let's keep it out of this PR.

TeXitoi:
I've also removed it on structopt. Big cache is too slow for Travis, and rust generate big cache. It can be a bit better if you do very custom cache selection, but, for example in nightly, it's always worst as the cache is only used one day.

I'll remove the commit in the hour.

therealprof:
I just checked Travis and I see what you mean. Indeed the Travis caching implementation for Rust is not ideal (e.g. it would make sense to remove old versions of the dependencies before storing the cache) so it grows rapidly and thus causing a slow down. OTOH removing it completely is not ideal either as it may cause crates.io overload to fail compilation.

Is just removing the cache line actually enough? Shouldn't it be explicitly set to false?

TeXitoi:
According to the timing, removing the line is enough. We can also purge the cache in the Travis interface.

TeXitoi:
So, should I remove the commit?

therealprof:
I've already purged the cache. I'm fine with disabling the cache but please remove the commit from this PR and let's do that separately.

TeXitoi:
removed

src/rcc.rs
```diff
@@ -96,43 +104,61 @@ impl CFGR {
...
-                    .plln()
-                    .bits(plln as u16)
-                    .pllp()
-                    .bits(pllp as u8)
-                    .pllsrc()
-                    .bit(self.hse.is_some())
-            });
-
-            (true, sysclk)
+        let pllq = (vco_in * plln) / 48_000_001 + 1;
```
therealprof: Why 48_000_001?

TeXitoi: Else you have to check for the case with exact division

src/rcc.rs
```diff
@@ -341,28 +371,36 @@ impl CFGR {
...
+
+        clocks
     }
 }
 
 /// Frozen clock frequencies
 ///
 /// The existence of this value indicates that the clock configuration can no longer be changed
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug)]
```
therealprof: Not a big fan of uneccessary `Debug` derives in embedded. They have the nasty tendency to massively blow up the generated code by accident.

TeXitoi: There is a dedicated commit on this ;-)

src/rcc.rs
TeXitoi: (i.e. we want to round upward, and adding 1 to the target division allow to round upward without test with a +1)

TeXitoi:
Debug removed

src/rcc.rs
Disasm: Counterexample: `96_000_001 // 48_000_001 + 1 == 2`, should be 3
I think this should be something like `(A + N - 1) / N`

TeXitoi: That's so frustrating to not have the possibility to make unit tests :-(

TeXitoi: fixed thanks!

src/rcc.rs
TeXitoi: Removed

TeXitoi:
@therealprof @Disasm Anything blocking on this? I think I've fixed every remarks.

src/rcc.rs
```diff
@@ -395,4 +433,18 @@ impl Clocks {
...
         self.sysclk
     }
+
+    /// Returns the frequency of the PLL48 clock line
+    pub fn pll48clk(&self) -> Option<Hertz> {
+        self.pll48clk
+    }
+
+    /// Returns true if the PLL48 clock is within USB
+    /// specifications. It is required to use the USB functionnality.
```
Disasm:
```suggestion
    /// specifications. It is required to use the USB functionality.
```

Disasm:
I'd rather added an example for one of the Nucleo boards, I believe they are more common than "generic STM32f401 development board". Other alternatives:
1. Add a link to this board so it becomes less generic
2. Remove the code that uses an LED on PC13 to make this example more generic

Also an `[[example]]` section in `Cargo.toml` will not hurt.

therealprof:
If you could fix the last nit and are happy with the commit structure, I think this is good to go.

TeXitoi:
For the example, that's because that's the only stm32f4 that I have. I can change it to whatever, as you wish

TeXitoi:
@Disasm I don't get the [[example]] thing, what do you want?

src/rcc.rs
TeXitoi: done

TeXitoi:
For the example, I've removed the LED and put:
```rust
 //! Target board: generic STM32f401 development board 
```
at the beginning of the file

Only the example thing, but I don't really know what to do. The commit structure is OK for me. It can also be squashed in a unique commit, I don't really care.

Disasm:
>  @Disasm I don't get the [[example]] thing, what do you want?

@TeXitoi I mean something like this:
https://github.com/stm32-rs/stm32f0xx-hal/blob/master/Cargo.toml#L91-L93
Other examples to not require the `usb_fs` feature (and they are for `stm32f411` as far as I can see).

TeXitoi:
examples in cargo.toml done

PR #130, opened 2020-03-04
Add Rng interface and an example that uses it.

tstellanova:
Enables use of the RNG (random number generator) peripheral available on some stm32f4 chips. 

This PR includes an example openocd configuration for running the example.

### DIFF
--- a/examples/rng-display.rs
+++ b/examples/rng-display.rs
@@ -1,0 +1,105 @@
+//! Generate random numbers using the RNG peripheral and display the values.
+//! This example is specifically tuned to run correctly on the
+//! stm32f4-discovery board (model STM32F407G-DISC1)
+//! This example requires the `rt` feature to be enabled. For example:
+//!
+//! ```bash
+//! cargo run --release --features stm32f407,rt  --example rng-display
+//! ```
+//!
+//! Note that this example requires the `--release` build flag because it is too
+//! large to fit in the default `memory.x` file provided with this crate.
+
+#![no_std]
+#![no_main]
+
+use stm32f4xx_hal as hal;
+
+#[cfg(not(debug_assertions))]
+use panic_halt as _;
+#[cfg(debug_assertions)]
+use panic_semihosting as _;
+
+use cortex_m_rt::ExceptionFrame;
+use cortex_m_rt::{entry, exception};
+use embedded_graphics::{fonts::Font6x8, prelude::*};
+
+use ssd1306::{prelude::*, Builder as SSD1306Builder};
+
+use hal::{i2c::I2c, prelude::*, stm32};
+use rand_core::RngCore;
+
+use arrayvec::ArrayString;
+use core::fmt;
+
+// dimensions of SSD1306 OLED display known to work
+pub const SCREEN_WIDTH: i32 = 128;
+pub const SCREEN_HEIGHT: i32 = 64;
+pub const FONT_HEIGHT: i32 = 8;
+/// height of embedded font, in pixels
+pub const VCENTER_PIX: i32 = (SCREEN_HEIGHT - FONT_HEIGHT) / 2;
+pub const HINSET_PIX: i32 = 20;
+
+#[entry]
+fn main() -> ! {
+    if let (Some(dp), Some(cp)) = (
+        stm32::Peripherals::take(),
+        cortex_m::peripheral::Peripherals::take(),
+    ) {
+        // Set up the system clock.
+        let rcc = dp.RCC.constrain();
+
+        // Clock configuration is critical for RNG to work properly; otherwise
+        // RNG_SR CECS bit will constantly report an error (if RNG_CLK < HCLK/16)
+        // here we pick a simple clock configuration that ensures the pll48clk,
+        // from which RNG_CLK is derived, is about 48 MHz
+        let clocks = rcc
+            .cfgr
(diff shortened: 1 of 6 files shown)


tstellanova:
I'll look into the CI failures. Looks like it's failing on some processors that don't have RNG

src/rng.rs
```diff
@@ -0,0 +1,115 @@
...
+    SeedError,
+}
+
+
+pub trait RngExt {
+    fn constrain(self) -> Rng;
+}
+
+impl RngExt for RNG {
+
```
therealprof:
```suggestion
```

therealprof:
Thank you for this PR. It looks very good in general. I'm not too thrilled about the OpenOCD stuff, this really shouldn't be in here. I'll add a few more comments and nits inline.

examples/rng-display.rs
```diff
@@ -0,0 +1,102 @@
...
+//! Note that this example requires the `--release` build flag because it is too
+//! large to fit in the default `memory.x` file provided with this crate.
+
+#![no_std]
+#![no_main]
+
+extern crate cortex_m;
+extern crate cortex_m_rt as rt;
+extern crate panic_semihosting;
+extern crate stm32f4xx_hal as hal;
```
therealprof: Can we please make this edition 2018?

src/rng.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// Enable RNG_CLK and the RNG peripheral.
+    /// Note that clocks must already be configured
+    /// such that RNG_CLK is not less than 1/16 HCLK,
+    /// otherwise all reads of the RNG will return
+    /// a ClockError (CECS error)
+    fn constrain(self) -> Rng {
```
therealprof: We should probably take the clock configuration here and make sure the prerequisites for the clock speeds are met.

src/rng.rs
```diff
@@ -0,0 +1,115 @@
...
+                    Ok(val as $type)
+                }
+
+                /// Fills buffer with random values, or return error
+                fn fill(&mut self, buffer: &mut [$type]) -> Result<(), ErrorKind> {
+                    const BATCH_SIZE: usize = 4 / mem::size_of::<$type>();
+                    let mut i = 0_usize;
+                    while i < buffer.len() {
+                        let random_word = self.next()?;
+                        let bytes: [$type; BATCH_SIZE] = unsafe { mem::transmute(random_word) };
```
therealprof: Not too hot about the transmute here. Probably should find a safer way to do it.

src/rng.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl rng::Read for Rng {
+    type Error = ErrorKind;
+
+    fn read(&mut self, buffer: &mut [u8]) -> Result<(), Self::Error> {
+        self.fill(buffer)
+    }
+}
+
+macro_rules! rng_core {
```
therealprof: Can't we simply use an associated type here instead of repeating the impl with a macro?

src/rng.rs
TeXitoi: https://doc.rust-lang.org/beta/std/primitive.u64.html#method.to_ne_bytes

So:
```suggestion
                        let bytes = random_word.to_ne_bytes();
```

src/rng.rs
TeXitoi: I'm not sure I understand your proposition. Do you want this macro be used:
```rust
rng_core!(usize);
rng_core(u32);
rng_core(u16)
rng_code(u8); 
```

src/rng.rs
TeXitoi: Maybe a bit more complex, but `from_ne_bytes()` and `to_ne_bytes()` looks like good tools.

src/rng.rs
therealprof: No, I mean move as much as possible out of such a macro, e.g. by turning the substituted type into an associated type.

tstellanova:
Taking another direction here and implementing the no_std interface of `rand_core`...stay tuned.

therealprof:
@tstellanova Cool, looking forward to it.

therealprof:
@tstellanova Looks excellent. I'm going to try it right away. Meanwhile would you please mention this new feature in https://github.com/stm32-rs/stm32f4xx-hal/blob/master/CHANGELOG.md and make sure to run `rustfmt` on it?

examples/rng-display.rs
```diff
@@ -0,0 +1,106 @@
...
+#[cfg(debug_assertions)]
+use panic_semihosting as _;
+
+use cortex_m_rt::ExceptionFrame;
+use cortex_m_rt::{entry, exception};
+use embedded_graphics::{fonts::Font6x8, prelude::*};
+
+use ssd1306::{prelude::*, Builder as SSD1306Builder};
+
+use hal::rng::RngExt;
```
therealprof: Please move this to the prelude so everyone who already uses the prelude can just use it.

therealprof:
I've tested it and it works a treat for me.

@TeXitoi Any thoughts from your side?

tstellanova:
The prelude change failed CI, I think because not all chips have the RNG peripheral.  I can either remove from prelude (similar to usb_fs) or add cfg features  in the prelude.  Any preferences?

therealprof:
Can you squash the commits a little to get rid of the failing intermediate changes? Then I would say this is ready to go in.

Issue #106, opened 2019-07-21
Add Timer::stop


PvdBerg1998:
Please expose a way to stop a timer. Even better would be a way to init a timer without starting it.
The pause line, located at https://docs.rs/stm32f4xx-hal/0.5.0/src/stm32f4xx_hal/timer.rs.html#210-231 could be extracted if I'm correct. The stm32f1xx hal (originally by japaric) already does this: https://docs.rs/stm32f1xx-hal/0.3.0/src/stm32f1xx_hal/timer.rs.html#149.

MarcoIeni:
I noticed that the `Cancel` trait exists in the [timer](https://docs.rs/crate/embedded-hal/0.2.3/source/src/timer.rs) API. Maybe for this function we should implement it?

therealprof:
@MarcoIeni Sounds like a good idea. Want to give it a shot?

MarcoIeni:
Yes, I will try to work on this :)

MarcoIeni:
A way to stop a timer has been provided by #137 and #148.

> Even better would be a way to init a timer without starting it.

Should we do this? Otherwise we can close this issue.

PR #137, opened 2020-03-27
Implement Cancel trait for Timer<SYST>

MarcoIeni:
This partially solves #106, because only I implemented `cancel()` only for `Timer<SYST>`. I would like to implement `Timer<$TIM>` once this is approved.

The error `Disabled` is returned because the timer [documentation](https://docs.rs/crate/embedded-hal/0.2.3/source/src/timer.rs) says:

> An error will be returned if the countdown has already been canceled or was never started.

## example

I added an example that shows how `cancel()` works, but I have a question: why the example doesn't work if I remove `use embedded_hal::timer::Cancel;`? For example the `wait()` function works even if CountDown is not imported explicitly.


MarcoIeni:
The `disable_counter()` is also used [here](https://docs.rs/stm32f1xx-hal/0.5.3/src/stm32f1xx_hal/timer.rs.html#248)

therealprof:
Thanks for the PR.

> I added an example that shows how cancel() works, but I have a question: why the example doesn't work if I remove use embedded_hal::timer::Cancel;? For example the wait() function works even if CountDown is not imported explicitly.

Maybe `embedded_hal::timer::Cancel` is not in the prelude of `embedded-hal`? We can always add it to this crates prelude so at least that should work nicely.

Other than that looks good to me but I need to take a closer look.

therealprof:
@MarcoIeni Sorry for taking so long until I got it. Looks good to me but would you please add a CHANGELOG entry for that?

MarcoIeni:
on my machine formatting looks good, I have rust 1.43 and `cargo fmt --all -- --check` succeeds.
There is an error in the CI or am I missing something?

therealprof:
Not sure but this rustfmt decided that the order should be different. Looks good though.

PR #139, opened 2020-04-15
Add support for the sdio peripheral

jkristell:
Left to do
- [x] Add it to all supported devices
- [x] Pinmap variants
- [x] Add an example
- [x] Document limitation (only SDHC cards)
- [x] Verify power configuration


thalesfragoso:
Thanks for working on this. Could you change the initialization of the clocks to use bit-banding ? That way we can avoid race conditions. Here is an example:

https://github.com/thalesfragoso/stm32f4xx-hal/blob/8f4c9a6b8f2cde41841429fc5d7b6f1b068a3b7b/src/pwm.rs#L165-L172

jkristell:
Sure! Will do that in the next version

There will probably be a couple of iterations of this before I'm happy with the api and code. But I wanted to share this now.

therealprof:
> Any idea on how I can check what devices in the stm32f4 family that have the sdio module and their pin mappings?

Seems all of them have it though with some lower pincount packages you can't use it electrically.

There was an attempt to extract the STM32CubeMX information to get all the possible pinouts but I'm not sure what the latest is about that. Maybe ask on the Matrix channel?

jkristell:
> > Any idea on how I can check what devices in the stm32f4 family that have the sdio module and their pin mappings?
> 
> Seems all of them have it though with some lower pincount packages you can't use it electrically.
> 
> There was an attempt to extract the STM32CubeMX information to get all the possible pinouts but I'm not sure what the latest is about that. Maybe ask on the Matrix channel?

Downloaded all the datasheets and found the pins manually. For this device it wasn't that tedious.

jkristell:
And nice catch by the CI. The F410 doen't have the sdio module.

therealprof:
Yeah, I missed that, too.

therealprof:
Needs reformatting. ;)

jkristell:
Needs some reformatting and some thinking over what to do with all the various information from the card. I'm leaning on just leaving all the information that is not strictly needed for basic functionality unparsed.

jkristell:
According the datasheet the stm32f446 should have the sdio peripheral. I will look into it, my guess is that it's a svd issue. But for now, just disable it.

jkristell:
Changelog entry added. I think this is ready for merging now.

jkristell:
Or maybe not. I should probably make read/write block take a &mut self instead.

therealprof:
> Or maybe not. I should probably make read/write block take a &mut self instead.

I haven't had time to think about the implications.

Maybe we can get a another opinion and a fresh pair of eyes on it? @aurelj @TeXitoi

jkristell:
Another issue is that this implementation doesn't capture the SDIO device from the pac, and among other problems I can think of, it makes the Sdio Sync. That's probably not correct.

I could take the same approach as the Rng module and make the RegisterBlock a member of the struct. I'm open to suggestions.

TeXitoi:
I'll try to look at it today or tomorrow.

examples/sdio-usb-block.rs
```diff
@@ -0,0 +1,218 @@
...
+    let mut red_led = {
+        let mut led = gpioc.pc1.into_push_pull_output();
+        let _ = led.set_low().ok();
+        led
+    };
+
+    let mut sdio = {
+        let d0 = gpioc
+            .pc8
+            .into_push_pull_output()
```
TeXitoi, line 97: You can remove this line, no?

src/sdio.rs
```diff
@@ -0,0 +1,825 @@
...
+}
+
+#[derive(Debug, Copy, Clone)]
+pub enum CardVersion {
+    V2 = 2,
+}
+
+#[derive(Debug, Copy, Clone)]
+pub enum CardType {
+    /// Standard Capacity (< 2Gn)
```
TeXitoi, line 151: Gb

src/sdio.rs
```diff
@@ -0,0 +1,825 @@
...
+    pub sys_spec_version: u8,
+    pub max_bus_clk_frec: u8,
+    pub rd_block_en: u8,
+    // V2
+    pub device_size: u32,
+}
+
+#[derive(Debug, Default, Copy, Clone)]
+pub struct Status {
+    bus_width: u8,
```
TeXitoi, line 217: Doesn't all these fields should be pub?

examples/sdio-usb-block.rs
jkristell: You are right. They shouldn't be needed.

src/sdio.rs
jkristell: > Great work!
> 
> The example is a bit complicated to be followed as an example, but I suppose that's your test application.

Thanks!

Not really my test program, but when I finished the driver someone made a usb mass storage crate, and I couldn't resist using it. I agree it's a bit to complicated. I could make a simpler example with just status and block read and write and have the more complex examples in my feather-board crate.

jkristell:
Thanks!

Not really my test program, but when I finished the driver someone made a usb mass storage crate, and I couldn't resist using it. I agree it's a bit to complicated. I could make a simpler example with just status and block read and write and have the more complex examples in my feather-board crate.

jkristell:
Rebased on latest master to fix merge conflict.

src/sdio.rs
```diff
@@ -0,0 +1,825 @@
...
+    pub rca: u32, // Relative Card Address
+    pub cid: [u32; 4],
+    pub csd: [u32; 4],
+    pub scr: [u32; 2],
+    pub status: [u32; 16],
+}
+
+impl Sdio {
+    pub fn new<PINS: Pins>(sdio: SDIO, _pins: PINS) -> Self {
+        //let sdio = unsafe { &*SDIO::ptr() };
```
therealprof, line 243:
```suggestion
```

src/sdio.rs
```diff
@@ -0,0 +1,825 @@
...
+    pub ctype: CardType,
+    pub ocr: u32,
+    pub rca: u32, // Relative Card Address
+    pub cid: [u32; 4],
+    pub csd: [u32; 4],
+    pub scr: [u32; 2],
+    pub status: [u32; 16],
+}
+
+impl Sdio {
```
therealprof, line 241: Could use some docstrings here and at the top of the file.

src/sdio.rs
```diff
@@ -0,0 +1,825 @@
...
+            performance_move: ((self.status[2] & 0xFF00) >> 8) as u8,
+            allocation_units: ((self.status[2] & 0x00F0_0000) >> 20) as u8,
+            erase_size: (((self.status[2] & 0xFF00_0000) >> 16) | (self.status[3] & 0xFF)) as u16,
+            erase_timeout: ((self.status[3] & 0xFC00) >> 10) as u8,
+            erase_offset: ((self.status[3] & 0x0300) >> 8) as u8,
+        }
+    }
+}
+
+impl Cmd {
```
therealprof, line 761: The docstrints on the functions in here are quite inconsistent

jkristell:
Added some docstrings and some inline comments about the commands.

I'm working on patches to the PAC so that we can have better register field descriptions of the SDIO peripherals in a later version. That will make some of the code a bit easier to understand.

PR #142, opened 2020-04-27
Setup GitHub Actions

therealprof:


TeXitoi:
What's the advantages of github actions comparing to Travis?

therealprof:
> What's the advantages of github actions comparing to Travis?

Tons:
- No central queue for all projects in the same organisation
- More runners (haven't found the limit yet)
- Tons of actions which can be simply used instead of having to script everything
- Better integration, e.g. annotations: https://github.com/stm32-rs/stm32f4xx-hal/runs/624314361?check_suite_focus=true
- No weird third-party API synching problems

Hopefully we'll also have less obscure "cache" problems but only time will tell. For some reason this repo currently tests slower than on Travis, other repos are massively faster -- not sure why.

PR #144, opened 2020-04-28
Update to embedded-hal v1.0.0-alpha.1

maxekman:
I needed to have the crate updated to use the new embedded-hal traits for the development of the I2S proposal (https://github.com/rust-embedded/embedded-hal/pull/204).

I'm new to the Rust and Rust embedded eco systems and not familiar with how the updates are usually done to all the HAL crates under stm32-rs, but I'll be glad to provide this if helpful. I'll leave it open for comments and further work until embedded-hal has released its next version (1.0?).

- [ ] Return correct errors where possible. Needs work to understand where we can get statuses from the MCU etc.
- [x] Decide which error to use for functions that is not expected to fail. Suggestions?
- [ ] Handle ignored return values in a idiomatic way. Suggestions?

### DIFF
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -1,6 +1,8 @@
 //! Delays
 
 use cast::u32;
+use core::convert::Infallible;
+
 use cortex_m::peripheral::syst::SystClkSource;
 use cortex_m::peripheral::SYST;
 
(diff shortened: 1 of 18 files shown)


src/timer.rs
```diff
@@ -105,7 +107,7 @@ impl Timer<SYST> {
     {
         syst.set_clock_source(SystClkSource::Core);
         let mut timer = Timer { tim: syst, clocks };
-        timer.start(timeout);
+        let _ = timer.try_start(timeout);
```
maxekman, line 110: Not sure about this way to handle the unused Result, suggestions?

src/watchdog.rs
```diff
@@ -88,17 +89,24 @@ impl IndependentWatchdog {
 }
 
 impl WatchdogEnable for IndependentWatchdog {
+    type Error = Infallible;
```
maxekman, line 92: I used this as a placeholder for this Error but I'm not sure if it's true?

src/delay.rs
```diff
@@ -1,6 +1,8 @@
 //! Delays
 
 use cast::u32;
+use core::convert::Infallible;
```
maxekman, line 4: I'm pretty sure this is not the correct error to use for functions that can't fail (or which proper error detection has yet to be implemented). Any suggestion?

luojia65:
You may update to version `1.0.0-alpha.1` and continue with the work now

maxekman:
Thanks @luojia65, I’ll do that.

eldruin:
Linking to https://github.com/rust-embedded/embedded-hal/pull/177

src/delay.rs
ryankurte: this is the correct marker to use when defining infallible implementations (as in, `type Error = Infallible;` in a trait implementation), where consuming functions they need to be generic over an error type.

maxekman: Thanks for the input!

maxekman:
I'll close this as it's not updated according to the guidelines for how to support 0.2.x and 1.0.0-alpha: https://github.com/rust-embedded/embedded-hal#adding-support-for-an-embedded-hal--alpha-version-in-a-hal-implementation

PR #145, opened 2020-04-28
Draft: I2S HAL implementation (for embedded-hal v1.x)

maxekman:
This is a WIP and possible reference implementation for the embedded-hal proposal for a I2S interface: https://github.com/rust-embedded/embedded-hal/pull/204

Not ready for review, left here in the open for comments and opinions.

Builds on #144 to be compatible with the HAL proposal https://github.com/eldruin/embedded-hal/tree/i2s, which is based on the master branch of embedded-hal.

- [ ] Fully support all devices with the macro system.
- [ ] Support possible second I2SPPL clock source in some devices (I think).
- [ ] Calculate viable audio clock frequencies. A bit tricky because the calculation is spread out between the RCC and I2S setup, they both have mult and div settings that have to play together to achieve the correct final MCKL out.
- [ ] Provide example, can be copied from the stm32f407-disc crate when done.
- [ ] Configurable I2S standard (Philips, right, left etc) and bit depth.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,8 +48,7 @@ default-features = false
 version = "1.0.2"
 
 [dependencies.embedded-hal]
-features = ["unproven"]
-version = "0.2.3"
+version = "=1.0.0-alpha.1"
 
 [dev-dependencies]
 panic-semihosting = "0.5.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,12 @@ default-features = false
 version = "1.0.2"
 
 [dependencies.embedded-hal]
-version = "=1.0.0-alpha.1"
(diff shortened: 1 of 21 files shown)


YruamaLairba:
I written a message here ? where is it ? i was saying i'm interested by your i2s work.

why it doesn't progress anymore ? because the i2s clock abstraction is near impossible to design ?

maxekman:
@YruamaLairba Not sure where your other messages are, but I did see the email notification a while back. I have unfortunately not had time to work on this lately. I agree that the clock interface is tricky (or near impossible) but have not yet given it much thought. I have only worked on the actual HAL implementation, not the setup code.

maxekman:
I'll close this as @samcrow has merged I2S support. Thanks for finishing it!

PR #148, opened 2020-05-01
Implement Cancel trait for Timer<TIM>

MarcoIeni:
Solves #106.
There is code duplication with [timer-syst](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/examples/timer-syst.rs) in the example. I don't know if it can be avoided.


MarcoIeni:
Thank you for reviewing Pull requests at the speed of light :rocket:

PR #149, opened 2020-05-01
System time based on static systick isr counter

h2obrain:
I added this to the hal while playing with it. Maybe it fits the upstream :)


src/systick.rs
```diff
@@ -0,0 +1,109 @@
...
+
+// there can only be one!
+#[exception]
+fn SysTick() {
+    free(|_|unsafe {
+        if let Some(systime) = &mut SYSTIME {
+            systime.systick += 1;
+        }
+    });
+}
```
therealprof, line 108: That is going to be an issue. We really shouldn't make any assumptions about the runtime here.

h2obrain: So is it possible to wrap an interrupt without making assumptions about the ruintime?
I mean something better than adding a global `fn SystickISR() {..}`, which has to be called from the implementation isr.

h2obrain:
Maybe this could just be an example without changes to the hal?

therealprof:
> Maybe this could just be an example without changes to the hal?

That is definitely possible. I was actually thinking about implementing a general "tick" counter crate a while back; something that you would initialise and then feed with monotonic counter updates which could be used for all kinds of purposes, e.g. a delay implementation.

PR #150, opened 2020-05-01
Dwt: Delay and stopwatch

h2obrain:
Implements "dwt cycle counter"-based `use embedded_hal::blocking::delay::{DelayMs, DelayUs};`
Additionally you can use it to measure execution times of code fragments.

### UPDATE2
Feel free to comment on which function names should change!
```rust
use crate::hal::{
    dwt::{ClockDiff, DelayS, DwtExt, StopWatch},
    prelude::*,
    stm32,
};
// Constrain DWT and DCB
let dwt = cp.DWT.constrain(cp.DCB, clocks);
// Create a (copy-/cloneable) delay instance
let delay = dwt.delay();
// DelayMs/DelayUs from hal
delay.delay_us(1000);
delay.delay_ms(1);
// DelayS from dwt.rs
delay.delay_s(0.0000005f32);
delay.delay_s(0.0000005f64);
// Measure some closures timing
let t = dwt.measure(|| delay.delay_ms(10));
let _t: f64 = t.as_secs_f64();
let _t: f32 = t.as_secs_f32();
let _t: u64 = t.as_nanos();
let _t: u32 = t.as_ticks();
// General stopwatch
let mut times = [0u32; 5]; // max 4 laps
let mut sw = self.stopwatch(&mut times, clocks);
sw.lap();
delay.delay_ms(10);
sw.lap();
sw.lap();
// Get all the lap times
{
    let mut lap = 1;
    while let Some(lap_time) = sw.lap_time(lap) {
        let _t = lap_time.as_secs_f64();
        lap += 1;
    }
}
```
### Iterators
A draft for iterators, which seems to need more code than I would have guessed, you can find here ([dwt.rs](https://github.com/h2obrain/stm32f4xx-hal/blob/dwt-stopwatch-iter/src/dwt.rs), [dwt-blinky.rs](https://github.com/h2obrain/stm32f4xx-hal/blob/dwt-stopwatch-iter/examples/dwt-blinky.rs))

### ClockDuration with from-functions

A ClockDuration with interchangeable types would be cool to do stuff like `delay.delay(ClockDuration::from_secs_f64(0.1));` along with this `let cd = dwt.measure(do_stuff); .. ; delay.delay(cd);`

A "the are no stupid solutions" implementation of this can found here ([dwt.rs](https://github.com/h2obrain/stm32f4xx-hal/blob/dwt-lazy-duration/src/dwt.rs), [dwt-blinky.rs](https://github.com/h2obrain/stm32f4xx-hal/blob/dwt-lazy-duration/examples/dwt-blinky.rs))

### DIFF
--- a/examples/dwt-blinky.rs
+++ b/examples/dwt-blinky.rs
@@ -1,0 +1,71 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+// Halt on panic
+use crate::hal::{
+    dwt::{ClockDuration, DwtExt},
+    prelude::*,
+    stm32,
+};
+use cortex_m;
+use cortex_m_rt::entry;
+use panic_halt as _;
+use stm32f4xx_hal as hal;
+
+#[entry]
+fn main() -> ! {
+    if let (Some(dp), Some(cp)) = (
+        stm32::Peripherals::take(),
+        cortex_m::peripheral::Peripherals::take(),
+    ) {
+        // Set up the LEDs. On the STM32F429I-DISC[O1] they are connected to pin PG13/14.
+        let gpiog = dp.GPIOG.split();
+        let mut led1 = gpiog.pg13.into_push_pull_output();
+        let mut led2 = gpiog.pg14.into_push_pull_output();
+
+        // Set up the system clock. We want to run at 48MHz for this one.
+        let rcc = dp.RCC.constrain();
+        let clocks = rcc.cfgr.sysclk(48.mhz()).freeze();
+
+        // Create a delay abstraction based on DWT cycle counter
+        let dwt = cp.DWT.constrain(cp.DCB, clocks);
+        let mut delay = dwt.delay();
+
+        // Create a stopwatch for maximum 9 laps
+        // Note: it starts immediately
+        let mut lap_times = [0u32; 10];
+        let mut sw = dwt.stopwatch(&mut lap_times);
+        loop {
+            // On for 1s, off for 1s.
+            led1.set_high().unwrap();
+            led2.set_low().unwrap();
+            delay.delay_ms(1000_u32);
+            sw.lap();
+            led1.set_low().unwrap();
+            led2.set_high().unwrap();
+            delay.delay_ms(900_u32);
+            // Also you can measure with almost clock precision
+            let cd: ClockDuration = dwt.measure(|| delay.delay_ms(100_u32));
+            let _t: u32 = cd.as_ticks(); // Should return 48MHz * 0.1s as u32
+            let _t: f32 = cd.as_secs_f32(); // Should return ~0.1s as a f32
+            let _t: f64 = cd.as_secs_f64(); // Should return ~0.1s as a f64
+            let _t: u64 = cd.as_nanos(); // Should return 100000000ns as a u64
+            sw.lap();
+
+            // Get all the lap times
+            {
+                let mut lap = 1;
+                while let Some(lap_time) = sw.lap_time(lap) {
+                    let _t = lap_time.as_secs_f64();
+                    lap += 1;
+                }
+            }
+
+            // Reset stopwatch
+            sw.reset();
+        }
+    }
+
+    loop {}
+}
--- a/src/dwt.rs
+++ b/src/dwt.rs
@@ -1,0 +1,200 @@
+//! Debug and trace and stuff
+
+use crate::rcc::Clocks;
+use crate::time::Hertz;
+use cortex_m::peripheral::{DCB, DWT};
+use embedded_hal::blocking::delay::{DelayMs, DelayUs};
+
+pub trait DwtExt {
+    fn constrain(self, dcb: DCB, clocks: Clocks) -> Dwt;
+}
+impl DwtExt for DWT {
+    /// Enable trace unit and cycle counter
+    fn constrain(mut self, mut dcb: DCB, clocks: Clocks) -> Dwt {
+        dcb.enable_trace();
+        self.enable_cycle_counter();
+        Dwt {
+            dwt: self,
+            dcb,
+            clocks,
+        }
+    }
+}
+
+pub struct Dwt {
+    dwt: DWT,
+    dcb: DCB,
+    clocks: Clocks,
+}
+impl Dwt {
+    /// Release the dwt and dcb control
+    /// # Safety
+    /// All instances of Delay and StopWatch become invalid after this
+    pub unsafe fn release(self) -> (DWT, DCB) {
+        (self.dwt, self.dcb)
+    }
+    /// Create a delay instance
+    pub fn delay(&self) -> Delay {
+        Delay {
+            clock: self.clocks.hclk(),
+        }
+    }
+    /// Create a stopwatch instance
+    /// # Arguments
+    /// * `times` - Array which will be holding the timings in ticks (max laps == times.len()-1)
+    pub fn stopwatch<'i>(&self, times: &'i mut [u32]) -> StopWatch<'i> {
+        StopWatch::new(times, self.clocks.hclk())
+    }
+    /// Measure cycles it takes to execute f
+    pub fn measure<F: FnOnce()>(&self, f: F) -> ClockDuration {
+        let mut times: [u32; 2] = [0; 2];
+        let mut sw = self.stopwatch(&mut times);
+        f();
+        sw.lap().lap_time(1).unwrap()
+    }
+}
+
+#[derive(Clone, Copy)]
+pub struct Delay {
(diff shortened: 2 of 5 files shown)


src/dwt.rs
```diff
@@ -0,0 +1,198 @@
+//! Debug and trace and stuff
+
+use cortex_m::peripheral::{DWT,DCB};
+use crate::rcc::Clocks;
+use crate::time::Hertz;
+//use core::u32;
+
+use embedded_hal::blocking::delay::{DelayMs, DelayUs};
+
+/* TODO only on arch7(e)m */
```
therealprof, line 10: Not going to be an issue here. STM32F4 will always be armv7em.

h2obrain: Ups. I forgot to delete those lines. I ported the from my c-codebase :)

h2obrain:
Thanks for the fast review!
- I will clean up the code asap
- The reason i chose f64 over f32 was because of f32s bad resolution (only ~6 digits), which looked kind of dangerous for a precision timer (`np.float32(1.000001)-np.float32(1)` = 9.536743e-07). But yeah, it's quite the overhead. I am new to rust so I am not sure, but in c I would always go for double, because that is was libc uses in printf.

therealprof:
Rust will use whatever is available natively and emulate the emulate the rest in software. But you only pay for what you use, so no f64 in code means no emulation of it. If you provide an API for f32 and f64 then the user could choose whether to pay for the overhead. 6 digits resolution seems good enough for most usecases to me, for higher resolution timing (e.g. on a instruction level) one would probably use a debug probe to do external tracing.

However now I'm wondering why the "native" format is not just clock cycles in an integer format with the option to internally calculate a time.

h2obrain:
Meh. Sorry about that.

examples/dwt-blinky.rs
```diff
@@ -0,0 +1,53 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+// Halt on panic
+#[allow(unused_extern_crates)] // NOTE(allow) bug rust-lang/rust#53964
+extern crate panic_halt; // panic handler
```
therealprof, line 7:
```suggestion
use panic_halt as _;
```

src/dwt.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+    )*}
+}
+impl_DelayFloatT!(for f32,f64);
+
+/// Very simple stopwatch
+/// Notes:
+/// - Max tdiff is < (1<<32) / HCLK
+/// - Assumes DWT cycle counter is enabled
+/// - Assumes Debug is set up
```
therealprof, line 142: Is that correct and what does that mean?

h2obrain: - The 32-bit cycle counter will overflow very soon
(eg. after ~25s@168MHz)
- Instanciating StopWatch does not enable
    - the cycle counter (`DWT.enable_cycle_counter()`)
    - nor the trace unit (`DCB.enable_trace()`)
    - it just assumes the `dcb.constrain(..)` stuff was done before

src/dwt.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub fn delay(duration: ClockDuration) {
+        let cycles = duration.ticks as u64;
+        Delay::delay_(DWT::get_cycle_count(), cycles);
+    }
+    fn delay_us_(self, us: u64) {
+        let start = DWT::get_cycle_count();
+        let cycles = (us * self.clock.0 as u64) / 1_000_000;
+        Delay::delay_(start, cycles);
+    }
+    fn delay_ms_(self, ms: u64) {
```
therealprof, line 68: Why the trailing underscores? Can't you make those be just the regular `delay_ms` and `pub` and simply remove the macro specialisation for `u64` further down?

h2obrain:
Okay, I will add some doc-strings.
What do you think of the iter for StopWatch and from..-functions for ClockDuration?

therealprof:
> Okay, I will add some doc-strings.
> What do you think of the iter for StopWatch and from..-functions for ClockDuration?

Can you point to what you mean?

@TeXitoi @aurelj Any opinions? Looks almost ready to me.

h2obrain:
iter for StopWatch ([dwt.rs](https://github.com/h2obrain/stm32f4xx-hal/blob/dwt-stopwatch-iter/src/dwt.rs), [dwt-blinky.rs](https://github.com/h2obrain/stm32f4xx-hal/blob/dwt-stopwatch-iter/examples/dwt-blinky.rs))

from..-functions for ClockDuration ([dwt.rs](https://github.com/h2obrain/stm32f4xx-hal/blob/dwt-lazy-duration/src/dwt.rs), [dwt-blinky.rs](https://github.com/h2obrain/stm32f4xx-hal/blob/dwt-lazy-duration/examples/dwt-blinky.rs))

h2obrain:
I wrote some doc-strings

src/dwt.rs
```diff
@@ -0,0 +1,219 @@
...
+use cortex_m::peripheral::{DCB, DWT};
+use embedded_hal::blocking::delay::{DelayMs, DelayUs};
+
+pub trait DwtExt {
+    fn constrain(self, dcb: DCB, clocks: Clocks) -> Dwt;
+}
+impl DwtExt for DWT {
+    /// Enable trace unit and cycle counter
+    fn constrain(mut self, mut dcb: DCB, clocks: Clocks) -> Dwt {
+        cortex_m::interrupt::free(|_| {
```
TeXitoi, line 14: Why interrupt free if you have ownership of the registers?

src/dwt.rs
```diff
@@ -0,0 +1,219 @@
...
+            }
+        }
+        impl DelayMs<$t> for Delay {
+            fn delay_ms(&mut self, ms: $t) {
+                self.delay_ms(ms as u64);
+            }
+        }
+    )*}
+}
+impl_DelayIntT!(for usize,  u32, u16, u8);
```
TeXitoi, line 132: You can avoid this macro by impl<T: Into<u64>> (maybe not for usize)

src/dwt.rs
```diff
@@ -0,0 +1,219 @@
...
+        self.timei
+    }
+    /// Resets recorded laps to 0 and sets 0 offset
+    pub fn reset(&mut self) {
+        self.timei = 0;
+        self.times[0] = DWT::get_cycle_count();
+    }
+    /// Record a new lap
+    /// NOTE If lap count exceeds maximum, the last lap is updated
+    pub fn lap(&mut self) -> &Self {
```
TeXitoi, line 166: Why returning &Self (instead of nothing or &must Self)?

src/dwt.rs
```diff
@@ -0,0 +1,219 @@
...
+
+pub struct Dwt {
+    dwt: DWT,
+    dcb: DCB,
+    clocks: Clocks,
+}
+impl Dwt {
+    /// Release the dwt and dcb control
+    /// NOTE all instances of Delay and StopWatch become invalid after this
+    pub fn release(self) -> (DWT, DCB) {
```
TeXitoi, line 37: Yeah, a proper implementation could borrow Dwt, bit the api will become annoying. What about putting this function unsafe? Even if that's not really unsafe, should only cause deadlocks.

src/dwt.rs
h2obrain: ups.. I removed this now. It was just some copy-pasta (especially the enable fmc-clock on the next line)

src/dwt.rs
h2obrain: changed to impl into

src/dwt.rs
h2obrain: Changed to unsafe

src/dwt.rs
h2obrain: I return nothing here, so I can write `sw.lap().lap_time(1).unwrap()` (see `fn measure(..)`).
Fixed the return type to `&mut Self`.

h2obrain:
> Needs CHANGELOG entry. And would you please remove the DRAFT marker?
I added some note to the CHANGELOG.
Should I squash the commits?
Would you prefer an extra commit for the example?
I guess the iterator for StopWatch and .from..-functions for ClockDuration are off the table then?

therealprof:
> Should I squash the commits?
> Would you prefer an extra commit for the example?

I don't care either way. If you'd like to hide the sausage making feel free to reorganise it, otherwise we can merge as is.

> I guess the iterator for StopWatch and .from..-functions for ClockDuration are off the table then?

We can always add them another time.

h2obrain:
@therealprof Okay.. Sorry, apparently you need to review it again :p "Requested changes must be addressed to merge this pull request."

therealprof:
bors retry

PR #161, opened 2020-05-12
This is an advanced example of stopwatch-with-ssd1306-and-interrupts.rs

mcuby:
This example requires the `rt` feature to be enabled. For this example, you need stm32f429i-disco (development kit) and ssd1306 (SPI).  You could find more details about this example here: http://www.mcu.by/rust-embedded-stopwatch/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,10 +54,11 @@ panic-semihosting = "0.5.3"
 cortex-m-semihosting = "0.3.3"
 arrayvec = { version = "0.5.1", default-features = false }
 panic-halt = "0.2.0"
-ssd1306 = "0.2.6"
-embedded-graphics = "0.4.9"
+ssd1306 = "0.3.1"
+embedded-graphics = "0.6.2"
 usb-device = "0.2.5"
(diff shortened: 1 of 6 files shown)


therealprof:
Hi @mcuby, thanks for your contribution. I haven't looked too closely but how different is this from https://github.com/stm32-rs/stm32f4xx-hal/blob/master/examples/stopwatch-with-ssd1306-and-interrupts.rs?

mcuby:
Hi @therealprof. This example is different in that you can continue the stopwatch, reset the stopwatch using the user button, use LEDs to indicate the stopwatch, work with SPI ssd1306 (was I2c) and, in addition, I have added an analog clock. This video is how it works.
[video](http://www.mcu.by/wp-content/uploads/2020/05/0-02-05-df265eeccb5cf8a60f1f1fb769099fdf27857bddf508e37aaade3fc93b92f3e8_207666c7b8fa7b.mp4?_=1)

therealprof:
Excellent.

mcuby:
Yes, I'll do it

PR #163, opened 2020-05-19
Make SDIO bus frequency selectable

plorefice:
There are cases in which the bus cannot run at 24MHz, for example due to PCB layout issues.

This PR makes the bus frequency selection a part of the public API, allowing the user to run the SDIO bus at a lower frequency, if necessary.

### DIFF
--- a/src/sdio.rs
+++ b/src/sdio.rs
@@ -116,7 +116,6 @@ enum PowerCtrl {
 }
 
 /// Clock frequency of a SDIO bus.
-#[allow(dead_code)]
 pub enum ClockFreq {
     F24Mhz = 0,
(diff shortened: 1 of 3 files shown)


src/sdio.rs
```diff
@@ -115,8 +115,9 @@ enum PowerCtrl {
     On = 0b11,
 }
 
+/// Clock frequency of a SDIO bus.
 #[allow(dead_code)]
```
therealprof, line 119: I believe this can go now as well since the enum is pub.

plorefice:
@therealprof thanks for your review, I have applied the requested changes.

PR #165, opened 2020-05-19
Move power handling to its own function and add short delay to register update

jkristell:
Did a quick read of the Reference manual, and the ST hal driver looking for potential timing issues. Noticed that their driver have a delay, to make sure the time between disabling and enabling the clock is long enough. So I implemented that and moved the card power functionality to its own function.

The sdio registers are sensitive to consecutive writes (If I understand the Reference manual correctly) in relation to both SDIOCLK and PCLK2,  and shouldn't be written to repeatable within certain number of cycles. But as far as I can tell now we should be ok without adding additional delays in block read/write.

Made this a draft as I'm not sure my approach to using the clocks struct to create a 2 ms delay is correct.


jkristell:
> Looks good to me if you still want to have it. If you, would you please fix the merge conflict?

I forgot about this one. I can rebase and port this once the other sdio branch is merged.

jkristell:
All done now. Should be ready for merge.

PR #167, opened 2020-05-26
Added default transactional spi impl

ryankurte:
Added transactional SPI implementation (see: https://github.com/rust-embedded/embedded-hal/pull/191)

This also includes a switch from `Void` to `Infallible` to be compatible with the updated hal.

Depends on: https://github.com/stm32-rs/stm32f4xx-hal/pull/172
Blocked on: https://github.com/rust-embedded/embedded-hal/pull/191
**Notes: Remove patch prior-to-merge**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -123,3 +123,6 @@ required-features = ["rt", "stm32f429"]
 [[example]]
 name = "rng-display"
 required-features = ["rt", "stm32f407"]
+
+[patch.crates-io]
+embedded-hal = { git = "https://github.com/ryankurte/embedded-hal.git", branch = "feature/spi-transactions" }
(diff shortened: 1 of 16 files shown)


Cargo.toml
```diff
@@ -124,3 +124,8 @@ required-features = ["rt", "stm32f429"]
 [[example]]
 name = "rng-display"
 required-features = ["rt", "stm32f407"]
+
+
+[patch.crates-io]
+#embedded-hal = { git = "https://github.com/ryankurte/embedded-hal.git", branch = "feature/spi-transactions" }
+embedded-hal = { path = "../embedded-hal" }
```
therealprof, line 131: Uhm... :D

ryankurte: yeeah i keep having to change things to try appease the cursed https://travis-ci.org/github/rust-embedded/embedded-hal/builds/691542832

PR #169, opened 2020-06-11
Move I2C write implementation into write_bytes function

therealprof:
This allows us to reuse the code for both write_read() and write() where
we need to add a STOP condition after sending the data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Make sure that I2C writes are concluded with a STOP condition
+
 ## [v0.8.2] - 2020-05-29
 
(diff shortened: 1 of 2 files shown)


Glaeqen:
I'd highlight that this PR fixes important fault in HAL: Now, `embedded_hal::blocking::i2c::Write` trait implementation satisfies contract regarding STOP condition at the end of the transmission. Previously it was missing.

I hope it'll get merged into mainline asap :)

therealprof:
Excellent, thanks for the confirmation.

therealprof:
Let me know if you need a release with it.

Glaeqen:
Well, I need this change all the time. Currently I'm using this HAL from cloned repo but if this change gets into crates.io, I'll obviousely switch. I don't know how troublesome is deploying new version of HAL with upgraded semver into crates.io. I don't want to bother you.

PR #170, opened 2020-06-14
Add implementation for Fastmode+ I2C available on select MCUs

therealprof:
This peripheral is very similar to the one found in newer MCUs e.g.
STM32F0 and was taken from there.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -1,4 +1,5 @@
 use core::ops::Deref;
+use embedded_hal::blocking::i2c::{Read, Write, WriteRead};
 
 use crate::stm32::i2c1;
 
@@ -41,8 +42,6 @@ use crate::stm32::I2C3;
     feature = "stm32f479"
 ))]
 use crate::stm32::{I2C1, I2C2, RCC};
-
-use embedded_hal::blocking::i2c::{Read, Write, WriteRead};
 
 #[cfg(any(
     feature = "stm32f401",
@@ -193,6 +192,13 @@ use crate::time::{Hertz, KiloHertz, U32Ext};
 
 /// I2C abstraction
 pub struct I2c<I2C, PINS> {
+    i2c: I2C,
+    pins: PINS,
+}
+
+#[cfg(any(feature = "stm32f413", feature = "stm32f423",))]
+/// I2C FastMode+ abstraction
+pub struct FMPI2c<I2C, PINS> {
     i2c: I2C,
     pins: PINS,
 }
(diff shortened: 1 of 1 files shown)


therealprof:
@aurelj @TeXitoi Any opinions on this?

aurelj:
> @aurelj @TeXitoi Any opinions on this?

I haven't looked in details but this seems great overall.

The FMPI2C peripheral definitions in the SVDs are quite messy and inconsistent, so I've tried to clean this up here:
  https://github.com/stm32-rs/stm32-rs/pull/390
This should allow to support FMPI2C on the f410, f412, f413, f423 and f446 very easily.

therealprof:
Wow, that is excellent! I've wanted to implement it for all STM32F4 that support it but ran into those gnarly inconsistencies in naming so it stripped it back to the only ones which compile with the current state.

I've successfully tested it with the STM32F413H Discovery board but only with regular I2C speeds (400kHz).

src/i2c.rs
```diff
@@ -617,6 +671,28 @@ impl<PINS> I2c<I2C3, PINS> {
...
+impl<PINS> FMPI2c<FMPI2C, PINS> {
+    pub fn fmpi2c(i2c: FMPI2C, pins: PINS, speed: KiloHertz, clocks: Clocks) -> Self
+    where
+        PINS: Pins<FMPI2C>,
+    {
+        // NOTE(unsafe) This executes only during initialisation
+        let rcc = unsafe { &(*RCC::ptr()) };
+
+        // Enable clock for FMPI2C
+        rcc.apb1enr.modify(|_, w| w.i2c4en().set_bit());
```
TeXitoi, line 684: Isn't it racy?

src/i2c.rs
```diff
@@ -852,3 +928,240 @@ where
...
+            scldel = 4;
+        } else {
+            presc = 0;
+            scll = cmp::max((((freq >> presc) >> 1) / speed.0) - 1, 255) as u8;
+            sclh = scll - 6;
+            sdadel = 1;
+            scldel = 3;
+        }
+
+        // Enable I2C signal generator, and configure I2C for 400KHz full speed
```
TeXitoi, line 967: Inconsistent comment?

therealprof: Yes, thanks for noticing. Guess I'll need to fix it in stm32f0xx-hal, too. ;)

src/i2c.rs
therealprof: Yes it is, any suggestions?

TeXitoi: Critical section, bit banging or &mut Rcc as parameter?

therealprof: Yeah, the usual suspects. ;)

At least the race is consistent with the already existing other I2C impls. I guess we might want to fix them all in the same way in one swoop?

TeXitoi: As you wish, there is a lot of racy code in this crate. Maybe we should open issues to track them?

therealprof: Let's do that.

therealprof: I don't have the time right now to check all the places which require attention but I've opened #171 to keep track of required fixes.

PR #172, opened 2020-06-18
Refactor for e-h v1.0.0-alpha

ryankurte:
Blocked on e-h v1.0.0 release (https://github.com/rust-embedded/embedded-hal/issues/177)

cc. @therealprof

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -131,7 +131,7 @@ impl Timer<SYST> {
 
 impl CountDown for Timer<SYST> {
     type Time = Hertz;
-    type Error = nb::Error<Error>;
+    type Error = Error;
 
     fn try_start<T>(&mut self, timeout: T) -> Result<(), Self::Error>
     where
@@ -160,7 +160,7 @@ impl CountDown for Timer<SYST> {
 impl Cancel for Timer<SYST> {
     fn try_cancel(&mut self) -> Result<(), Self::Error> {
         if !self.tim.is_counter_enabled() {
-            return Err(nb::Error::Other(Error::Disabled));
+            return Err(Error::Disabled);
         }
 
         self.tim.disable_counter();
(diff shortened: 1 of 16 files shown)


src/timer.rs
```diff
@@ -132,8 +131,9 @@ impl Timer<SYST> {
 
 impl CountDown for Timer<SYST> {
     type Time = Hertz;
+    type Error = nb::Error<Error>;
```
eldruin, line 134: Does this need to be `nb::Error<Error>`? I think could just be `Error`.

ryankurte: the `try_wait` function below returns `nb::Error::WouldBlock` so, without changing this, yes?

ryankurte: wait you’re right i think i have this backwards, will try fixing it tomorrow. thanks for the comment!

ryankurte: got it, thanks for the review!

maxekman:
Hi! How does this compare to my v1.0.0-alpha.1 branch #144? Can we merge/split the effort? I mainly want to be compatible with 1.0.x for the development of #145.

ryankurte:
i somehow missed your PR when i made this one, they should be functionally equivalent. i haven't diffed em to check but, happy to close this in favour of yours and add the delta if there is anything missing.

maxekman:
Sure, if that works for you @ryankurte. I'll see if I can finish it soon.

Do you think it could be useful to have a v1.0.0-dev branch for this repo so we can both merge any changes to it?

maxekman:
There is now some new info about how to support both 0.2.x and 1.0.0-alpha: https://github.com/rust-embedded/embedded-hal#adding-support-for-an-embedded-hal--alpha-version-in-a-hal-implementation

ryankurte:
while it's possible to do this it's also pretty easy to build adaptors between versions, and to update things, so i definitely would not think it an _expectation_ that both are supported.

burrbull:
Closed in favor of #388

PR #174, opened 2020-06-20
Implement ExtiPin for pins in output mode

mvirkkunen:
Since `InputPin` is implemented for pins in output mode, it also makes sense to implement `ExtiPin`. Especially useful in open drain mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Reexport PAC as `pac` for consistency with other crates, consider `stm32` virtually deprecated
+- Added external interrupt (EXTI) support for output pins
 
 ## [v0.8.3] - 2020-06-12
 
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -103,22 +103,22 @@ macro_rules! exti_erased {
                         syscfg.exticr1.modify(|r, w| unsafe {
                             w.bits((r.bits() & !(0xf << offset)) | ($extigpionr << offset))
                         });
-                    },
+                    }
                     4..=7 => {
                         syscfg.exticr2.modify(|r, w| unsafe {
                             w.bits((r.bits() & !(0xf << offset)) | ($extigpionr << offset))
                         });
-                    },
+                    }
                     8..=11 => {
                         syscfg.exticr3.modify(|r, w| unsafe {
                             w.bits((r.bits() & !(0xf << offset)) | ($extigpionr << offset))
                         });
-                    },
+                    }
                     12..=15 => {
                         syscfg.exticr4.modify(|r, w| unsafe {
                             w.bits((r.bits() & !(0xf << offset)) | ($extigpionr << offset))
                         });
-                    },
+                    }
                     _ => {}
                 }
             }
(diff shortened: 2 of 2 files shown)


therealprof:
I was thinking of revisiting the any output pin can be an input pin implementation. I only noticed that detail yesterday and I think it is dangerous and wrong to allow reading from non-OD output pins.

Also I have my doubts that EXTI on output pins can work (at least reliably) at all. The RM for the F411 clearly says (emphasis mine):

> External interrupt/wakeup lines
> 
> All ports have external interrupt capability. To use external interrupt lines, **the port must be configured in input mode**, refer to Section 10.2: External interrupt/event controller (EXTI) and Section 10.2.3: Wakeup event management.

mvirkkunen:
I thought that it was a bit odd too that non-open drain pins were also inputs! I suppose reading the value should be well defined though since the RM says the input circuitry is also enabled in output mode though so it's not *wrong* per se.

Somehow I missed that bit about no interrupts on output lines (I didn't immediately find a mention of that in the EXTI section itself when I checked earlier), but wouldn't that also cover open drain? Open drain is classified under "output" in the manual. I hope that doesn't mean you can't do interrupts on OD pins. That would be a major limitation for any bidirectional OD bus.

therealprof:
> I thought that it was a bit odd too that non-open drain pins were also inputs! I suppose reading the value should be well defined though since the RM says the input circuitry is also enabled in output mode though so it's not _wrong_ per se.

Yes, it is not wrong to do this, it just doesn't make a lot of sense and it's easy to misinterpret the API as being autoswitching.

> Somehow I missed that bit about no interrupts on output lines (I didn't immediately find a mention of that in the EXTI section itself when I checked earlier), but wouldn't that also cover open drain?

Well, it's possible that the documentation is wrong. 😅

> Open drain is classified under "output" in the manual. I hope that doesn't mean you can't do interrupts on OD pins.

I think that's what it means.

mvirkkunen:
Did some cursory research, F1 and F3 have the same wording as F4 in the RM. Some F0 and G0 chips on the other hand have the wording changed to "To use external interrupt lines, the given pin must not be configured in analog mode or being used as oscillator pin, so the input trigger is kept enabled." I'm beginning to wonder if that's what it's supposed to say for all chips. But I guess there's no way to know for sure what each chip is supposed to support if we don't trust the RM.

mvirkkunen:
I actually went and asked ST about it and this is what they had to say

>my colleague confirmed to me that the EXTI should work in all the mode where GPIO IDR registers give valid information. It would not work in the analog mode. Only thing is that this functionality is not tested and that is the reason why it is not the reference manual.

I think the interrupts should work fine, at least I don't see any reason why they shouldn't. The input circuitry is connected in output mode too and the EXTI lines should be connected to that.

It's probably a good idea to remove the interrupt support from purely output pins (i.e. not open drain) because using them in that mode makes no sense. It should probably be a separate PR though as it's theoretically a breaking change.

therealprof:
Thanks for the confirmation. What does this mean for this PR?

Also I'm not too hot about the introduction of more macros, as they make everything even more unreadable. Is there no way we could use generic impls instead?

mvirkkunen:
My interpretation is that the feature I'm proposing to add should be OK to add.

Personally I find a non-exported macro nicer than copying and pasting 60 lines of code to change one word in it. After looking at some other code just now I found out that you can nest `macro_rules` though so it could be made a tiny bit simpler (less arguments).

All the HALs are kind of a mess of macros right now, and even then the "type erased" GPIO bits have similar repeated bits. I think all the pins could and should be defined as something like type aliases for a generic type with const generics (once those land), but that's also a story for another PR.

therealprof:
> My interpretation is that the feature I'm proposing to add should be OK to add.

Yes, I just wasn't sure about this part: 

> It's probably a good idea to remove the interrupt support from purely output pins (i.e. not open drain) because using them in that mode makes no sense.

Isn't that what you're freshly adding e.g. here:

```
            exti_erased!($PXx<Input<MODE>>, $extigpionr);
```

> Personally I find a non-exported macro nicer than copying and pasting 60 lines of code to change one word in it.

I fully agree but that's not what I meant. I mean we should try to come up with pure code means to implement features rather than repeating code hundred of times, both explicitly by verbatim repetition and implicitly via macros.

> All the HALs are kind of a mess of macros right now, and even then the "type erased" GPIO bits have similar repeated bits. I think all the pins could and should be defined as something like type aliases for a generic type with const generics (once those land), but that's also a story for another PR.

A lot of the current woes would go away with const generics and/or GATs.

mvirkkunen:
Heh, I think I got a bit mixed up there. What I meant was that EXTI should probably be implemented for exactly the same things `InputPin` is implemented for (barring hardware restrictions). Currently `InputPin` is also implemented for push-pull outputs, but it probably should be changed so it's only implemented for OD outputs, and *that* is the breaking change.

Would you like me to change this PR to only add EXTI support for open drain outputs?

therealprof:
If it works for any output type than this should be fine. Glad we could clarify this. ;)

therealprof:
It is missing a CHANGELOG.md entry though. ;)

mvirkkunen:
Added the entry. I have another EXTI related PR lined up after this as well, but didn't send them at the same time because they mess with the same code and would conflict.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Reexport PAC as `pac` for consistency with other crates, consider `stm32` virtually deprecated
+- Added external interrupt (EXTI) support for output pins to allow using interrupts with open drain
```
therealprof, line 13: Uhm, but you're adding the support for all output modes, right? 😅

mvirkkunen: I added that bit to at least try to explain why somebody would want to use interrupts with "output" pins. If you want I can reword it.

I think it'd be better to just remove `InputPin` as well as EXTI from non-OD pins in a future release though.

therealprof: Well, if you add EXTI support for any output mode then that's what the CHANGELOG entry should say.

mvirkkunen: Now it only says that.

PR #178, opened 2020-07-03
Update synopsys-usb-otg dependency to v0.2.0

Disasm:


Disasm:
Note that this is a breaking change.

Disasm:
Oops, forgot about this. Fixed!

PR #179, opened 2020-07-03
Move sdio response parsing to library

jkristell:
Clean up the sdio driver by moving the parsing of the various sd response to a separate library https://crates.io/crates/sdio-host. Some other cleanups related to this change was made as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ nb = "0.1.2"
 rand_core = "0.5.1"
 stm32f4 = "0.11"
 synopsys-usb-otg = { version = "0.2.0", features = ["cortex-m"], optional = true }
+sdio-host = { version = "0.2.0", optional = true }
 
 [dependencies.bare-metal]
 version = "0.2.5"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -32,6 +32,7 @@ nb = "0.1.2"
 rand_core = "0.5.1"
 stm32f4 = "0.11"
 synopsys-usb-otg = { version = "0.1.0", features = ["cortex-m"], optional = true }
+sdio-host = { path = "../sdio-host" }
```
therealprof, line 35: This will obviously not work. ;)

jkristell: I'm probably the only user of the driver anyway, so if I just put it behind a feature flag :-)

therealprof: > I think it's a good idea to have some of the code in a library that could be shared with other host driver implementations, and make this driver a bit less cluttered.

I fully agree.

jkristell: I think the library is ready for a first release, but the naming issue is holding me back. I'm not really sure of the scope of the library, right now It's just SD memory card stuff, but I guess most controller are capable of sdio, so it could potentially (but unlikely, I don't think there are that many sdio devices developed anymore?) be extended with data structures for sdio as well. Something with the words  sd/sdio/sdhost - host/proto/datatypes ... arghh :)

therealprof: > I'm probably the only user of the driver anyway, so if I just put it behind a feature flag :-)

Maybe not a bad idea to do that similarly to the USB stuff.

jkristell:
> Looks mostly good to me, maybe feature gating SDIO would be a good thing to not have an additional dependency for non-users.
> 
> Also this is missing the CHANGELOG entry.

Added a sdio feature gate and Changelog Entry.

PR #182, opened 2020-07-17
Add a qei example

bschwind:
I tested this on a ["black pill"](https://stm32-base.org/boards/STM32F411CEU6-WeAct-Black-Pill-V2.0) dev board. This example just lights up the LED based on the rotary encoder direction, though the code also calculates the difference in position as you rotate it, so you could run some other code based on that difference.

On a side note, what are peoples' setups for building/flashing/running these examples? My setup:

* MacOS
* [`STM32F411CEU6`-based board](https://stm32-base.org/boards/STM32F411CEU6-WeAct-Black-Pill-V2.0)
* `CP2102` USB-serial converter <-> `STM32F411`
  * `TX` <-> `A10`
  * `RX` <-> `A9`
  * `3V3` <-> `3.3`
  * `GND` <-> `GND`
* `stm32flash` CLI for flashing the firmware
  * I have to hold the `BOOT0` button down and then press the `NRST` button to put it in flashing mode
* [`serial-monitor`](https://crates.io/crates/serial-monitor) for identifying the correct serial-port path and monitoring serial output

Some other examples are using `cortex_m_semihosting::{hprint, hprintln};` for output. Is that only for the ST-link device? Is it worth switching to that over serial?

Pics with the working QEI example:

IMG_0065
IMG_0066

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,3 +127,7 @@ required-features = ["rt", "stm32f429"]
 [[example]]
 name = "rng-display"
 required-features = ["rt", "stm32f407"]
+
+[[example]]
+name = "qei"
+required-features = ["rt", "stm32f411"]
--- a/examples/qei.rs
+++ b/examples/qei.rs
@@ -1,0 +1,67 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+// Demonstrates the use of a rotary encoder. This example was tested
+// on a "black pill" USB C board:
+// https://stm32-base.org/boards/STM32F411CEU6-WeAct-Black-Pill-V2.0
+//
+// The rotary encoder A and B pins are connected to pins A0 and A1,
+// and they each have a 10K ohm pull-up resistor.
+
+// Halt on panic
+use panic_halt as _;
+
+use cortex_m;
+use cortex_m_rt::entry;
+use embedded_hal::Direction as RotaryDirection;
+use stm32f4xx_hal::{delay::Delay, prelude::*, qei::Qei, stm32};
+
+#[entry]
+fn main() -> ! {
+    let dp = stm32::Peripherals::take().expect("Failed to get stm32 peripherals");
+    let cp = cortex_m::peripheral::Peripherals::take().expect("Failed to get cortex_m peripherals");
+
+    // Set up the LED. This is pin C13 on the "black pill" USB C board here:
+    // https://stm32-base.org/boards/STM32F411CEU6-WeAct-Black-Pill-V2.0
+    let gpioc = dp.GPIOC.split();
+    let mut led = gpioc.pc13.into_push_pull_output();
+
+    // Set up the system clock.
+    let rcc = dp.RCC.constrain();
+    let clocks = rcc.cfgr.freeze();
+
+    // Create a delay abstraction based on SysTick.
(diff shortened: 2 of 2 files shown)


bschwind:
> Please add an entry to Cargo.toml so it'll get built by CI.

Sure thing! Should I just add `stm32f411` as the feature since I tested it on that board? I imagine this works on every F4 board though...

therealprof:
> On a side note, what are peoples' setups for building/flashing/running these examples? My setup:

Most people will use some debugger and `probe-rs` for flashing and everything else.

> Some other examples are using cortex_m_semihosting::{hprint, hprintln}; for output. Is that only for the ST-link device? Is it worth switching to that over serial?

No, but it requires a debugger to be connected and attached which can be a pain in the neck. Please don't use semihosting.

therealprof:
> Sure thing! Should I just add `stm32f411` as the feature since I tested it on that board? I imagine this works on every F4 board though...

It may or may not. Feel free to try but adding a specific board is certainly a safe bet and fine for me.

bschwind:
@therealprof what debugger do you use? I'll have to try out `probe-rs`

therealprof:
I have like 10 different debuggers here. 😅 Various onboard ST-Link variants, ST-Link V2 clone, an ST-Link V3/Set, my own custom board with DAP42, onboard JLink, various onboard DAPLink implementations, KitProg v2 and v3...

A DAPLink implementation is most open, fast and versatile for `probe-rs`. ST-Link is probably the easiest and cheapest to obtain but not the fastest and v3 only works with STM chips. Followed by J-Link which is more versatile but proprietary and thus can be annoying to use.

bschwind:
> A DAPLink implementation is most open, fast and versatile for `probe-rs`

Sounds great, I'll check that one out. Curious about your custom board with DAP42, is it an open design? I could get the parts and assemble a PCB if it is and you recommend it over other debuggers :)

bschwind:
Also, sorry, one last question: If you don't recommend using semihosting for debug output, what would you recommend? Plain serial?

therealprof:
My DAP42 is basically just a breakout board with a few additional components, but there's also PCB designs available, e.g.:
https://github.com/devanlai/dap42-hardware
or from our Rust embedded team members:
https://github.com/korken89/micro-dap
https://github.com/adamgreig/ffp (even comes with custom CMSIS-DAP firmware; haven't tried it yet but it is supposed to be great!)

therealprof:
Either plain serial or if the prerequisite of a debugger to get information out is okay: https://crates.io/crates/rtt-target

eldruin:
I could flash a CMSIS-DAP firmware also on an ST-Link V2 clone. It was the only thing that allowed me to unlock a PineTime (nrf52).

therealprof:
> I could flash a CMSIS-DAP firmware also on an ST-Link V2 clone. It was the only thing that allowed me to unlock a PineTime (nrf52).

Which one did you use? I also tried DAP42 on the STM32F103 but with mixed success...

eldruin:
Just [this](https://mvdlande.wordpress.com/2015/10/05/cmsis-dap-on-a-cheap-st-link-v2-mini-adapter/). Pretty raw though :)

therealprof:
Yeah, that is very raw.

eldruin:
It did not break my USB port though :D

PR #185, opened 2020-07-19
Add DmaExt::split() and minimal dma::Stream.

m-ou-se:
Very minimal implementation of DMA. Splits DMA1 or DMA2 into their 8 streams, which give access to the stream's registers and interrupt bits.

Can be extended later, but this should be a good start.

### DIFF
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -1,0 +1,185 @@
+use core::marker::PhantomData;
+
+use crate::stm32::{DMA1, DMA2, RCC};
+
+pub trait DmaExt: Sized {
+    /// Enables the DMA peripheral and splits it into streams.
+    fn split(
+        self,
+    ) -> (
+        Stream0<Self>,
+        Stream1<Self>,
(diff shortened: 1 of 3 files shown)


therealprof:
@m-ou-se Looks great to me, but I'm a bit puzzled how to use this as is. Would it be possible to come up with a simple example?

m-ou-se:
`.st()` gives access to the [`dma::ST`](https://docs.rs/stm32f4/0.11.0/stm32f4/stm32f413/dma1/struct.ST.html) struct, the register bank of the dma stream. So using it isn't simple yet, it's almost the same as if using the `stm32f4` crate directly. The only useful this things adds is allowing `.split()`ting the DMA structs to to hand different dma streams to different parts of the code, just like `dp.GPIOA.split()`.

Adding a nice interface for setting all the settings and starting transfers etc. can be done separately in a later PR. Those would simply be additions to the `Stream` `impl`s.

PR #189, opened 2020-07-21
Cleanups and bugfixes for the sdio driver

jkristell:
This updates the driver to use the latest sdio-host release and has some refactoring of the datapath code. It also fixes some stability issues with sd card writes that I was seeing.

With this patch my [usb mass storage example](https://github.com/jkristell/feather-f405/blob/master/examples/sd-usb-mass-storage.rs) works reliable.

So this is a pretty important patch for the sdio driver.


jkristell:
This is ready for review/merge now @therealprof

PR #193, opened 2020-07-28
 Allow DMA on both halves of a split Serial<USART> separately.

bobmcwhirter:
Based upon and dependent upon https://github.com/stm32-rs/stm32f4xx-hal/pull/186

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -1471,7 +1471,7 @@ macro_rules! halUsartImpl {
                                     .dmat().enabled()
                             })
                         }
-                        _ => {}
+                        DmaConfig::None => {}
                     }
 
                     Ok(Serial { usart, pins }.config_stop(config))
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -374,6 +374,8 @@ use crate::gpio::AF11;
 use crate::gpio::{Alternate, AF7, AF8};
 use crate::rcc::Clocks;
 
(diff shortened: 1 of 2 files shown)


therealprof:
@bobmcwhirter Mind rebasing this so we can have a look at the gist of this change?

bobmcwhirter:
Will do shortly.

bobmcwhirter:
rebased/repushed.

thalesfragoso:
Hmm, seems like the first commit is undoing one of my last changes.

bobmcwhirter:
I probably screwed up my rebase I guess. Will fix Monday.

bobmcwhirter:
Hopefully this looks better.

src/serial.rs
```diff
@@ -1442,6 +1453,27 @@ macro_rules! halUsartImpl {
...
+                            })
+                        }
+                        DmaConfig::TxRx => {
+                            usart.cr3.write(|w| {
+                                w
+                                    .dmar().enabled()
+                                    .dmat().enabled()
+                            })
+                        }
+                        _ => {}
```
thalesfragoso, line 1474: Maybe this should be `DmaConfig::None` instead ?

bobmcwhirter: Ah, good point. I'll modify.

wrt hardware, yes, on an stm32f401re against my USART6.

PR #194, opened 2020-07-30
Basic SDRAM FMC driver

iostat:
Title says it all. This is a basic driver allowing one to use the FMC peripheral to drive an SDRAM module which can then be accessed directly in the memory space of the MCU. It /should/ work on F427, F429, F437, and F439 devices. However, I only have a STM32F429I-Discovery board to test it against. Mostly ported from the STM32 HAL/LL libraries. Included is a simple RAM test in `examples/` targeting the STM32F429I-Discovery board

It'd be nice if this could be expanded to support NOR/PSRAM and all the other goodies that the FMC offers, but I don't have any hardware to test it with. I believe the 429I-EVAL includes a NOR Flash and SDRAM on a single bus, so that may be interesting to ultimately target.

One particular divergence from the usual patterns in this module is that we `Peripherals::steal()` to get access to the GPIOs to set up their AF/Speed/Pullup. This is mainly due to the sheer number of pins that would otherwise have to be passed by the user, which seemed inconvenient at best. The majority of the FMC pins aren't swappable, so it made the most sense ergonomically to just manipulate them directly using `steal()`. The few pins which /are/ swappable (and whose usage of swapping affects behavior) are captured in the type system.


therealprof:
Thanks for the PR. I was planning to add FSMC support at some point and since they're very similar we should probably try to add them in one go and in a consolidated way.

I need to look at your proposal at more detail but I'll add a few things I've noticed right away.

Cargo.toml
```diff
@@ -88,6 +88,8 @@ usb_hs = ["synopsys-usb-otg", "synopsys-usb-otg/hs"]
 
 sdio = ["sdio-host"]
 
+fmc = []
```
therealprof, line 91: Pleasy only add meta-features if absolutely required. The reason to have those for usb and sdio is because that's the only way to use optional external dependencies.

examples/stm32f429idisco-fmc.rs
```diff
@@ -0,0 +1,163 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+// This example is designed to be run on the STM32F429I-DISCOVERY board
+
+// Halt on panic
+#[allow(unused_extern_crates)] // NOTE(allow) bug rust-lang/rust#53964
+extern crate panic_halt; // panic handler
```
therealprof, line 9:
```suggestion
use panic_halt as _; // panic handler
```

examples/stm32f429idisco-fmc.rs
```diff
@@ -0,0 +1,163 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+// This example is designed to be run on the STM32F429I-DISCOVERY board
+
+// Halt on panic
+#[allow(unused_extern_crates)] // NOTE(allow) bug rust-lang/rust#53964
```
therealprof, line 8:
```suggestion
```

src/fmc.rs
```diff
@@ -0,0 +1,564 @@
...
+    SDCMR, SDCR, SDRTR, SDSR, SDTR,
+};
+
+type FMCAF = crate::gpio::Alternate<crate::gpio::AF12>;
+macro_rules! fmc_gpio {
+    ($gpioX:ident.$pXY:ident) => {
+        $gpioX
+            .$pXY
+            .into_alternate_af12()
+            .set_speed(crate::gpio::Speed::High)
```
therealprof, line 19: Why only `High` instead of `VeryHigh`?

src/fmc.rs
```diff
@@ -0,0 +1,564 @@
...
+}
+
+impl SDRAMConfig {
+    fn configure_gpio<CCE: SDRAMClockAndChipEnable, SDNWE: PinSDNWE>(
+        &self,
+        cce: CCE,
+        sdnwe: SDNWE,
+    ) {
+        let peripherals = unsafe { crate::stm32::Peripherals::steal() };
+        let gpiod = peripherals.GPIOD.split();
```
therealprof, line 223: Those operations are risky as they do RMW operations. At the very least we need to somehow protect these. Also I think those GPIOs don't belong into something called `SDRAMConfig` as the GPIOs are completely independent of what exactly you're attaching to your parallel bus.

Probably better to move the GPIO configuration out into it's own thing. Also an interesting approach to explore would be to have the traditional `new` function where you have to pass in all the correct pins and a (unsafe) `new_unchecked` function which assumes that the application has properly set up everything.

src/fmc.rs
```diff
@@ -0,0 +1,564 @@
...
+        fmc_gpio!(gpiof.pf3); // PF3 -> A3
+        fmc_gpio!(gpiof.pf4); // PF4 -> A4
+        fmc_gpio!(gpiof.pf5); // PF5 -> A5
+        fmc_gpio!(gpiof.pf12); // PF12 -> A6
+        fmc_gpio!(gpiof.pf13); // PF13 -> A7
+        fmc_gpio!(gpiof.pf14); // PF14 -> A8
+        fmc_gpio!(gpiof.pf15); // PF15 -> A9
+        fmc_gpio!(gpiog.pg0); // PG0  -> A10
+
+        // always configure pin for address lines A0-A10 (minimum address size is 11-bit)
```
therealprof, line 242: Shouldn't that comment be over the actual configuration block?

src/fmc.rs
iostat: Primarily because it's unnecessary. Using a higher `Speed` means using more GPIO drivers (more drive strength), which means more potential EMI emissions and power consumption (best practice is to use the lowest Speed suitable for the application). The FMC runs the memory at half of HCLK, which puts you in the ballpark of 90MHz `SDCLK` frequency on a maxed out F429 running at 180MHz. Speed::High the lowest setting which gives us the bandwidth we need.

src/fmc.rs
iostat: >Those operations are risky as they do RMW operations. At the very least we need to somehow protect these. 

Agree with the risk, I was hoping it's mitigated by the fact that the operations they do are idempotent (i.e., they never disable any GPIO peripherals or pins, only additively enable them)

>Also I think those GPIOs don't belong into something called SDRAMConfig as the GPIOs are completely independent of what exactly you're attaching to your parallel bus. Probably better to move the GPIO configuration out into it's own thing.

Do you mean the GPIOs which are passed into `configure`, or generally the notion of `impl SDRAMConfig` being ultimately responsible for setting up the GPIOs. If the former -- those GPIOs are necessary for SDRAM functionality. The are passed into `SDRAMConfig.configure(...)` which consumes the SDRAMConfig and returns an SDRAM. This is deliberate done so that at least ownership of the swappable SDRAM control pins is taken in this case. If the latter, it's only done that way to separate out the logic of the different steps (i.e., so we don't have a gigantic `configure()` function)

>Also an interesting approach to explore would be to have the traditional new function where you have to pass in all the correct pins and a (unsafe) new_unchecked function which assumes that the application has properly set up everything.

A `new` function was my initial approach, but was abandoned due to the potential combinatorial explosion in future support for all device families and FMC configurations (i.e., `device family <-> gpio ports mappings/AFs <-> possible FMC configurations w/r/t bus widths and other pins`). OTOH, I can totally see a separate `FMCAddressBus`, `FMCDataBus`, etc. set of traits which /could/ encapsulate the different non-swappable pins, but again, their construction would involve significant boilerplate and doesn't seem to me to bring much value unless the rest of FMC's more sophisticated capabilities are supported (e.g., NOR/PSRAM mux, etc.)

iostat:
Thanks for taking a look at this and giving valuable feedback! I've replied to a few of these with my motivations/thought process, and will fix the remainder later tonight. Happy to keep discussing and refining this until its something everyone's OK with! I originally wanted to kickstart a discussion of how an FMC API would work as in issue, but it seemed more fruitful to start with an MVP and refine it with something that can be tested for ergonomics and functionality -- hence the PR.

therealprof:
@iostat I absolutely prefer the MVP approach!

src/fmc.rs
therealprof: > Agree with the risk, I was hoping it's mitigated by the fact that the operations they do are idempotent (i.e., they never disable any GPIO peripherals or pins, only additively enable them), and banking on the fact that the RegisterBlocks are guarded by VolatileCells in the upstream API

VolatileCell is not providing any protection against racy RMW operations.

> Do you mean the GPIOs which are passed into configure, or generally the notion of impl SDRAMConfig being ultimately responsible for setting up the GPIOs.

Both. The pins used by FMC/FSMC are the same, independent of the mode or devices connected to the bus. Probably we should even separate the bus setup altogether from the configuration of the channels.

> A new function was my initial approach, but was abandoned due to the potential combinatorial explosion in future support for all device families and FMC configurations (i.e., device family <-> gpio ports mappings/AFs <-> possible FMC configurations w/r/t bus widths and other pins).

It's not too bad, really. We'll have to deal with that anyway at some point.

I can test e.g. with a STM32F413H Disco (FSMC) and would certainly like to see the display work with this driver. I do have it running (in a hackish way) and can certainly provide some details about that.

iostat: > VolatileCell is not providing any protection against racy RMW operations.

Someone would be using this code with ownership of the peripheral register blocks anyway when doing their GPIO setup, and the VolatileCell guards against races due to the compiler reordering the accesses. I suspect we might not be on the same page with what "racy" means, could you elaborate?

> Both. The pins used by FMC/FSMC are the same, independent of the mode or devices connected to the bus. Probably we should even separate the bus setup altogether from the configuration of the channels.

I just started doing this, but this does mean a separate type for each possible width of address bus (+3 types for data buses). To enforce stuff like valid values of Column vs Row bits, how do you feel about requiring `const_generics` to support this?

therealprof: > Someone would be using this code with ownership of the peripheral register blocks anyway when doing their GPIO setup

How do you mean? You're conjuring the PeripheralBlocks out of thin air with your `steal()`. Stealing != owning.

> I just started doing this, but this does mean a separate type for each possible width of address bus (+3 types for data buses).

Not necessarily. Depending on how we want to achieve that we could do a lot of different things, including automatically generating enums with all possible variants or specififying the maximum possible and allowing dummy pins.

> how do you feel about requiring const_generics to support that?

Anything not requiring nightly is fair game. 😅

iostat: > How do you mean? You're conjuring the PeripheralBlocks out of thin air with your steal(). Stealing != owning.

You still need ownership of PeripheralBlocks to make the SCKEx, SDNEx, and SDNWE pins that you need in order to call `SDRAMConfig::configure`. So yes, on the surface I'm conjuring it up, but in practice, it's ultimately pointing the same instance of PeripheralBlocks that the caller owns anyway in order to be able to kick off those steals in the first place. Which is why I'm not sure how a race how would happen. 

> including automatically generating enums with all possible variants or specififying the maximum possible and allowing 

This sounds like we'd need proc macros, which I'm not against all, and might be the way forward as long as anything stable is fair game

therealprof: > You still need ownership of PeripheralBlocks to make the SCKEx, SDNEx, and SDNWE pins that you need in order to call SDRAMConfig::configure. So yes, on the surface I'm conjuring it up, but in practice, it's ultimately pointing the same instance of PeripheralBlocks that the caller owns anyway in order to be able to kick off those steals in the first place. Which is why I'm not sure how a race how would happen.

Not sure how you guarantee that everything is pointing to the same RegisterBlock, the pins could be very well on different ports. But that point is moot anyway because you're not proving anything at all by having access to a few pins because the registers are shared on purpose between multiple GPIOs which means that individual register access to stateful registers need to be guarded.

> This sounds like we'd need proc macros, which I'm not against at all, and might be the way forward as long as anything stable is fair game

You probably mean macros in general? Procedural macros would be a first in an HAL impl.  😅

iostat: > Not sure how you guarantee that everything is pointing to the same RegisterBlock, the pins could be very well on different ports. But that point is moot anyway because you're not proving anything at all by having access to a few pins because the registers are shared on purpose between multiple GPIOs which means that individual register access to stateful registers need to be guarded.

I can prove it by reading the fine source code :P . I'm well aware it's a weak argument, I'm mostly just making a case for developer convenience at the cost of bending some absolute (general) guarantees which seem to hold anyway in this particular context.

> You probably mean macros in general?

I actually mean procedural macros. You'd need to do some subset of dependent types to get, for example, the number of row and column bits vs. address bus size, which you can approximate by implementing a subset of something like [Haskell's `singletons`](https://cs.brynmawr.edu/~rae/papers/2012/singletons/paper.pdf). (full disclosure, i'm one of *those* people :P). I suspect you'd need to end up needing to manipulate the AST in ways more sophisticated than `macro_rules` lets you do. Naturally, all that wouldn't affect anything at runtime, just there to let the typechecker keep you honest

> Procedural macros would be a first in an HAL impl

Sounds great for marketing :P

therealprof: > I can prove it by reading the fine source code :P . I'm well aware it's a weak argument, I'm mostly just making a case for developer convenience at the cost of bending some absolute (general) guarantees which seem to hold anyway in this particular context.

I don't see how you can claim there're no concurrent register accesses from reading the code. When creating such interfaces it is important to balance the "convenience" of not having to specify a boatload of specific pins against the possibility of incorrect use. I actually prefer the verbosity of the type system over the lack of any sanity checking (which is exactly what the C world is suffering quite a bit from).

It seems that other folks are strongly considering two versions, the "traditional" one with the verbosity and a `new_unchecked` which expects the user to setup everything correctly before going ahead and simply use it. I'd be open to such a split approach.

> I actually mean procedural macros.

I'm not sure we really need procedural macros but I'm also not opposed to using them. I'm just afraid it's going to be tedious due to the requirement to have the proc macros in a separate crate.

iostat: > I don't see how you can claim there're no concurrent register accesses from reading the code.

I'm think I'm starting to see what you mean by races. I can think of only one instance where the non-atomicity of RMW would be an issue here -- an interrupt firing as we're in the middle of the setup process with someone's handler changing a pin on one of the GPIO peripherals we use to a High-Z state or something like that. Since this code is intended to be used as part of initialization, would it be acceptable to mitigate it by disabling interrupts at the start of `configure` and reenabling them before returning? We could also forgo stealing the FMC peripheral as it's wholly unnecessary to steal it -- and its only been stolen cause we had stolen access to the peripheral block anyway. 

> When creating such interfaces it is important to balance the "convenience" of not having to specify a boatload of specific pins against the possibility of incorrect use.

With you 100%. I just also want to note, much of this conversation for me is to see where the balance lies here, not just to defend my approach wholesale (though, admittedly, I may be just a tad biased :P)

richardeoin:
I've started [a crate](https://github.com/stm32-rs/stm32-fmc) to provide FMC/FSMC drivers that can be used for all the stm32 parts. It currently only supports some F7, some H7, and F469/F479, and also only has SDRAM support so far.

Would you be interested in combining this PR into the crate? I think some of the tasks would be:

* Looking at the [register defintions](https://github.com/stm32-rs/stm32-fmc/blob/master/src/ral/peripherals/fmc.rs) and deciding if they apply to F427/F429/F437/F439, otherwise creating another file.
* Creating a [memory device definition](https://github.com/stm32-rs/stm32-fmc/tree/master/src/devices) for the SDRAM on the STM32F429I-Discovery
* Finally creating an [implementation](https://docs.rs/stm32-fmc/0.1.0/stm32_fmc/#implementation-guide) here in the F4 HAL

Let me know your thoughts. I will be adding more support for the H7, but happy to help here as well.

src/fmc.rs
iostat: So, I just made a variant of this PR where you pass in all pins, (including address and data buses), and it seems to not actually fit in the 64K flash defined in `memory.x` for all the examples. Building with `--release` seems to make it fit, as does bumping the `FLASH = LENGTH 64K` to `128K` in `memory.x`, but this is a mere workaround for the bloat in compiled code that is introduced by this approach (i.e., forcing developers to use `--release` to use this may hamper their efforts in debugging, which seems unfavorable to me). I'm also not entirely sure where it originates, as not even `#[inline]`ing seems to resolve it. With all of that in mind, I'm starting to be convinced that stolen Address/Data buses and owned `FMC` and other support pins might be the only viable approach.

I defer to you to determine the best way forward here. Worth mentioning is after this is ironed out I'm planning to make a similar driver for the F429's `LTDC` peripheral which has a similar requirement with a wide GPIO bus necessary for its functionality (albeit, there are more pinswapping options, so it may require large numbers of traits, etc. which could create similar code bloat issues). All that to say, whatever is the approach taken here will likely end up informing that effort. Perhaps it makes more sense to divert this effort into @richardeoin's crate, so that building the examples isn't bounded by the presence of the `fmc` submodule here?

therealprof: `#[inline]` has no effect in dev mode. Need to go `#[inline(always)]` for that.

iostat: Just peppered #[inline(always)] everywhere, same result.

iostat: I've pushed an [`fmc-support-owned`](https://github.com/iostat/stm32f4xx-hal/tree/fmc-support-owned) branch if you want to see if you can make it fit. I'm going to make this PR rely less on unsafe in the interim w/r/t owning FMC and taking ownership of as many pins as possible.

therealprof: Need to go deeper into the rabbit hole and figure out where the bloat is coming from. It's unfortunate that dev mode is pretty much unusable for embedded but it's also not worth working around it in weird ways  because it's quite likely that you'll hit the next road block right around the corner (usually in the form of code executing so slow due a massive amount of silly function calls that all essential timing goals are missed).

therealprof: > I've pushed an [`fmc-support-owned`](https://github.com/iostat/stm32f4xx-hal/tree/fmc-support-owned) branch if you want to see if you can make it fit.

Making code work in dev mode is a tertiary goal for me at best. As I mentioned before I'd be okay with having two ways of initialising, one "regular" `new()` with all pins and one `new_unchecked()`, e.g. like done here: https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/spi.rs. Though I will insist on that being `unsafe` because I very much disagree on the `unchecked` postfix providing a clear enough message about the real unsafety (and that `unsafe` is only applicable to memory safety) of this.

burrbull:
Closing in favor of stm32-fmc.

PR #195, opened 2020-08-01
Sdio driver updates

jkristell:
Various cleanups, 
The biggest change is that Cmd and command related functionality moved to the sdio-host crate.


jkristell:
Ready for review / merge

PR #203, opened 2020-08-15
Adding NACK detection and fault condition clearing

### LINKED ISSUE #202: i2c write hangs when receiving NACK
The current I2C implementation will hang indefinitely whenever a NACK is received on the bus during a write request.

This occurs because the current implementation waits for the `SR1.addr` bit to be set before determining that the address transfer is complete. The RM notes that this bit does not assert when a NACK is received.

image

ryan-summers:
This PR fixes #202 by adding the following:
1. During all busy-wait I2C loops, error flags are continually checked.
2. All fault flags have been updated to manually clear the fault indication.

cc: @therealprof

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -550,6 +550,10 @@ impl PinScl<FMPI2C> for PF15<AlternateOD<AF4>> {}
 pub enum Error {
     OVERRUN,
     NACK,
+    TIMEOUT,
+    BUS,
+    CRC,
+    ARBITRATION,
 }
 
 #[cfg(any(
@@ -767,6 +771,53 @@ where
         self.i2c.cr1.modify(|_, w| w.pe().set_bit());
     }
 
+    fn check_and_clear_error_flags(&self) -> Result<(), Error> {
+        let sr1 = self.i2c.sr1.read();
+
+        // Clear all pending error flags. We have already read the SR1, so it's safe to clear them
+        // before returning the error code.
+        self.i2c.sr1.modify(|_, w| {
+            w.timeout()
+                .clear_bit()
+                .pecerr()
+                .clear_bit()
+                .ovr()
+                .clear_bit()
+                .af()
+                .clear_bit()
+                .arlo()
+                .clear_bit()
+                .berr()
+                .clear_bit()
+        });
+
+        if sr1.timeout().bit_is_set() {
+            return Err(Error::TIMEOUT);
+        }
+
+        if sr1.pecerr().bit_is_set() {
+            return Err(Error::CRC);
+        }
+
+        if sr1.ovr().bit_is_set() {
+            return Err(Error::OVERRUN);
+        }
+
+        if sr1.af().bit_is_set() {
+            return Err(Error::NACK);
+        }
+
+        if sr1.arlo().bit_is_set() {
+            return Err(Error::ARBITRATION);
+        }
+
+        if sr1.berr().bit_is_set() {
+            return Err(Error::BUS);
+        }
+
+        Ok(())
+    }
+
     pub fn release(self) -> (I2C, PINS) {
         (self.i2c, self.pins)
     }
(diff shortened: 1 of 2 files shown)


src/i2c.rs
```diff
@@ -803,7 +803,18 @@ where
             .write(|w| unsafe { w.bits(u32::from(addr) << 1) });
 
         // Wait until address was sent
-        while self.i2c.sr1.read().addr().bit_is_clear() {}
+        while {
+            let sr1 = self.i2c.sr1.read();
+
+            // If we received a NACK, then this is an error
+            if sr1.af().bit_is_set() {
+                self.i2c.sr1.modify(|_, w| w.af().clear_bit());
```
therealprof, line 811: Would it make sense to clear all errors here instead of just the NACK condition?

src/i2c.rs
```diff
@@ -803,7 +803,18 @@ where
...
+                return Err(Error::NACK);
+            }
+
+            // Wait for the address to be acknowledged.
```
therealprof, line 815:
```suggestion
            // Wait for the address to be acknowledged
```

src/i2c.rs
```diff
@@ -830,6 +841,7 @@ where
 
             // If we received a NACK, then this is an error
             if sr1.af().bit_is_set() {
+                self.i2c.sr1.modify(|_, w| w.af().clear_bit());
```
therealprof, line 844: Same question as above. If guess we should also factor this out into its own function to make it clearer what it does.

src/i2c.rs
```diff
@@ -767,6 +771,53 @@ where
         self.i2c.cr1.modify(|_, w| w.pe().set_bit());
     }
 
+    fn check_and_clear_error_flags(&self) -> Result<(), Error> {
+        let sr1 = self.i2c.sr1.read();
+
+        // Clear all pending error flags. We have already read the SR1, so it's safe to clear them
+        // before returning the error code.
+        self.i2c.sr1.modify(|_, w| {
```
ryan-summers, line 779: It appears that modifying this register before checking the bits below causes the below checks to never succeed.

src/i2c.rs
ryan-summers: This should now be implemented in a `check_and_clear_errors()` function.

src/i2c.rs
ryan-summers: What's you're preferred commenting grammar? I can update any new comments, but I'm unsure how to format them based on this change request.

src/i2c.rs
ryan-summers: Should now be refactored into `check_and_clear_errors()`

ryan-summers:
Changelog should now be added. 

On an STM32F407-DISC1, I can confirm that attempting to transact with a device that is not on the bus causes an `Error::NACK` to be returned.

src/i2c.rs
therealprof: Well, for me full sentences or sentences in a paragraph end on a period. Sentence-like remarks don't. It's likely we're not consistent about that, it's also likely I don't always notice but when I do I tend to nitpick. ;)

src/i2c.rs
```diff
@@ -550,6 +550,10 @@ impl PinScl<FMPI2C> for PF15<AlternateOD<AF4>> {}
 pub enum Error {
     OVERRUN,
     NACK,
+    TIMEOUT,
+    BUS,
+    CRC,
+    ARBITRATION,
```
therealprof, line 556: Hm, I was about to move all errors over to `embedded-error`. But I can also do that independently.

therealprof:
Looks good to me in general. I'll have a closer look in a bit and will do a small regression check.

therealprof:
Okay, works fine. Do you want to do some cleanup or is this ready to go?

ryan-summers:
It should be good on my end unless you have any further desired changes

PR #204, opened 2020-08-18
Move maximum clock frequencies into public constants

therealprof:

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -29,7 +29,93 @@ pub struct Rcc {
     pub cfgr: CFGR,
 }
 
-const HSI: u32 = 16_000_000; // Hz
+/// Built-in high speed clock frequency
+pub const HSI: u32 = 16_000_000; // Hz
+
+#[cfg(any(
+    feature = "stm32f401",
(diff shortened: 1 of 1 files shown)


src/rcc.rs
```diff
@@ -470,9 +476,9 @@ impl Clocks {
     /// Returns true if the PLL48 clock is within USB
     /// specifications. It is required to use the USB functionality.
     pub fn is_pll48clk_valid(&self) -> bool {
-        // USB specification allow +-0.25%
+        // USB specification allows +-0.25%
         self.pll48clk
-            .map(|freq| (48_000_000 - freq.0 as i32).abs() <= 48_000_000 * 25 / 10000)
+            .map(|freq| (48_000_000 - freq.0 as i32).abs() <= 120_000)
```
TeXitoi, line 481: Why doing this change ? Better to explicit this value by adding to the above comment "i.e. +-120kHz"

therealprof: Because I find it more readable than the weird integer formula.

therealprof:
bors r=TeXitoi

PR #209, opened 2020-08-28
Allow for skipping an ongoing DMA transfer if not using double buffering

thalesfragoso:
This also fixes a problem that could happen when disabling the stream with an active transfer. The code is also updated to use the `embedded-dma` crate now.

As draft because I didn't test it on hardware yet.


thalesfragoso:
Ok, tested on hardware, everything seems to be working fine.

therealprof (approved):
Excellent, thanks a lot @thalesfragoso!

bors r+

PR #210, opened 2020-09-07
Update SPI implementation to more compliance version

no111u3:


therealprof (commented):
Thanks! Looks good on first glance; I want to run some tests though.

src/spi.rs
```diff
@@ -954,6 +954,64 @@ where
...
+
+    fn is_send_only(&mut self) -> bool {
+        self.spi.cr1.read().bidimode().bit_is_set() && self.spi.cr1.read().bidioe().bit_is_set()
+    }
+
+    fn set_bidi(&mut self) {
+        self.spi
+            .cr1
+            .modify(|_, w| w.bidimode().clear_bit().bidioe().clear_bit());
+    }
```
thalesfragoso, line 971: We should probably mark these methods as `inline(always)`.

src/spi.rs
```diff
@@ -954,6 +954,64 @@ where
...
+
+    fn read_u8(&mut self) -> u8 {
+        // NOTE(read_volatile) read only 1 byte (the svd2rust API only allows reading a half-word)
+        unsafe { ptr::read_volatile(&self.spi.dr as *const _ as *const u8) }
+    }
+
+    fn send_u8(&mut self, byte: u8) {
+        // NOTE(write_volatile) see note above
+        unsafe { ptr::write_volatile(&self.spi.dr as *const _ as *mut u8, byte) }
+    }
```
thalesfragoso, line 1013: We should probably mark these methods as `inline(always)`.

src/spi.rs
```diff
@@ -1014,48 +1072,63 @@ where
...
-        } else if sr.modf().bit_is_set() {
-            nb::Error::Other(Error::ModeFault)
-        } else if sr.crcerr().bit_is_set() {
-            nb::Error::Other(Error::Crc)
-        } else if sr.rxne().bit_is_set() {
-            // NOTE(read_volatile) read only 1 byte (the svd2rust API only allows
-            // reading a half-word)
-            return Ok(unsafe { ptr::read_volatile(&self.spi.dr as *const _ as *const u8) });
+        if !self.is_send_only() {
+            nb::block!(self.check_read())?;
```
thalesfragoso, line 1076: Do we really want to change behavior here ? Before this method could return `WouldBlock`.

src/spi.rs
```diff
@@ -1014,48 +1072,63 @@ where
...
         } else {
-            nb::Error::WouldBlock
-        })
+            Ok(0)
+        }
     }
 
     fn send(&mut self, byte: u8) -> nb::Result<(), Error> {
-        let sr = self.spi.sr.read();
+        nb::block!(self.check_send())?;
```
thalesfragoso, line 1084: Same thing as above.

src/spi.rs
```diff
@@ -954,6 +954,64 @@ where
...
     }
 
+    pub fn set_send_only(&mut self) {
+        self.spi
+            .cr1
+            .modify(|_, w| w.bidimode().set_bit().bidioe().set_bit());
+    }
+
+    fn is_send_only(&mut self) -> bool {
+        self.spi.cr1.read().bidimode().bit_is_set() && self.spi.cr1.read().bidioe().bit_is_set()
```
thalesfragoso, line 964: We probably want to read it only once, it's a volatile read so the compiler won't be able to optimize it.

src/spi.rs
```diff
@@ -954,6 +954,64 @@ where
...
+    }
+
+    fn set_bidi(&mut self) {
```
thalesfragoso, line 967: Shouldn't this be `pub` ?

src/spi.rs
```diff
@@ -1014,48 +1072,63 @@ where
...
+
+        // We have a 32 bit buffer to work with, so let's fill it before checking the status
+        for word in words {
+            self.send_u8(*word);
+        }
+
+        // Do one last status register check before continuing
+        self.check_send().ok();
+        Ok(())
+    }
```
thalesfragoso, line 1133: We probably want to check what was the mode before and return to it before returning from the method, i.e. if it was bidirectional we should go back to it.

thalesfragoso (commented):
Thanks for the PR, I just have some questions and suggestions. Also, a changelog entry is missing.

no111u3:
Closed by provide new version for this pr - #349.

PR #212, opened 2020-09-13
Draft: I2S HAL implementation (for embedded-hal v0.2.x)

maxekman:
This is a WIP and possible reference implementation for the embedded-hal proposal for a I2S interface: https://github.com/rust-embedded/embedded-hal/pull/204

Not ready for review, left here in the open for comments and opinions.

- [ ] Fully support all devices with the macro system.
- [ ] Support possible second I2SPPL clock source in some devices (I think).
- [ ] Calculate viable audio clock frequencies. A bit tricky because the calculation is spread out between the RCC and I2S setup, they both have mult and div settings that have to play together to achieve the correct final MCKL out.
- [ ] Provide example, can be copied from the stm32f407-disc crate when done.
- [ ] Configurable I2S standard (Philips, right, left etc) and bit depth.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,8 +48,12 @@ default-features = false
 version = "1.0.2"
 
 [dependencies.embedded-hal]
+# TODO: Update version when I2S proposal is finalized:
+# https://github.com/rust-embedded/embedded-hal/pull/204
 features = ["unproven"]
-version = "0.2.3"
+# version = "0.2.3"
+git = "https://github.com/eldruin/embedded-hal/"
+branch = "i2s-v0.2.x"
 
 [dev-dependencies]
 panic-semihosting = "0.5.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,8 +52,9 @@ version = "1.0.2"
 # https://github.com/rust-embedded/embedded-hal/pull/204
 features = ["unproven"]
 # version = "0.2.3"
-git = "https://github.com/eldruin/embedded-hal/"
-branch = "i2s-v0.2.x"
+# git = "https://github.com/eldruin/embedded-hal/"
+# branch = "i2s-v0.2.x"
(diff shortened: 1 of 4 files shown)


YruamaLairba:
ooops, i started to work on i2s without seeing there was already a work in progress. I will look at it.

I'm a bit skeptical about calculating i2s clock and audio frequency. As you said it's a nightmare to calculate it directly and i don't see how to design an appropriate HAL interface for that since the calculation is splitted between RCC and IS2 peripherals. Adding to this, such interface tend to hide differences between real and expected clock. 
Instead, it's easy for an user to do scripts to try all valid combination to find the ones that fulfill their requirement, So why not to  let use prescaler value ?

samcrow:
FYI: I recently made another I2S driver similar to this one.

Driver library: https://github.com/samcrow/stm32_i2s/
`i2s` module in a fork of stm32f4xx-hal: https://github.com/samcrow/stm32f4xx-hal/blob/i2s/src/i2s.rs

Some properties:
* Supports all four master/slave and transmit/receive modes
* Supports all for data formats (16 bits, 16 bits in 32-bit frames, 24 bits in 32-bit frames, 32 bits)
* Automatic clock division calculation based on the input frequency and desired sample rate (or the division can be specified manually)
* Not compatible with embedded-hal yet
* I have tested master transmit mode fairly thoroughly, but have not tested the other modes

I don't want to steal your thunder or invalidate the work you've already done. If you think any parts of my code are good, you're welcome to copy them.

maxekman:
@samcrow I'm glad it’s being worked on! Haven’t had time myself. Please also take anything you want from my work and submit a PR if you have something that is more complete. Maybe I’ll have time to revisit this some day, but not right now. Thanks for being so considerate!

maxekman:
I'll close this as @samcrow has merged I2S support. Thanks for finishing it!

Issue #171, opened 2020-06-16
Audit crate for potential racy code


therealprof:
As noted in #170 this crate has code in various places which is potentially racy, e.g. when using interrupt handlers due to the lack of e.g. synchronisation primitives.

Known areas which require a solution:

- [ ] I2C

TeXitoi:
Thanks for opening this.

PR #213, opened 2020-09-20
Use bitbanding to avoid data races

thalesfragoso:
Partially addresses #171.


TeXitoi (approved):
Great!

Maybe svd2rust should export something to ease this.

therealprof (approved):
Excellent, thanks you!

bors r+

PR #214, opened 2020-09-20
add USART example

nebelgrau77:


examples/serial.rs
```diff
@@ -0,0 +1,54 @@
+#![no_main]
+#![no_std]
+
+use panic_halt;
```
therealprof, line 4:
```suggestion
use panic_halt as _;
```

examples/serial.rs
```diff
@@ -0,0 +1,54 @@
...
+    prelude::*,
+    stm32,
+    serial::config::Config,
+    serial::Serial};
+
+use core::fmt::Write; // for pretty formatting of the serial output
+
+#[entry]
+fn main() -> ! {
+
```
therealprof, line 20:
```suggestion
```

therealprof (commented):
If I'm not mistaken this HAL impl should allow you to `writeln!` directly onto the `serial`. Splitting should only be necessary if you want to read and write in different parts of your application.

birkenfeld:
Why not show receiving data as well? That's usually the more tricky part...

therealprof:
@birkenfeld The more examples the merrier... ;) I don't see any problem with a simple example demonstrating a very common usecase (like printing logging info) and a more complex one doing more fancy stuff like interacting with an external device.

nebelgrau77 (commented):
Modified as suggested.

nebelgrau77:
> If I'm not mistaken this HAL impl should allow you to `writeln!` directly onto the `serial`. Splitting should only be necessary if you want to read and write in different parts of your application.

I guess my n00bness strikes again: I saw that in other examples and assumed you actually had to split. But then again, commenting the `split` line and replacing `tx` with `serial` (I guess you meant that) will not compile: 

```rust
writeln!(serial, "value: {}\r", value).unwrap();
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `stm32f4xx_hal::serial::Serial<stm32f4xx_hal::stm32::USART2, (stm32f4xx_hal::gpio::gpioa::PA2<stm32f4xx_hal::gpio::Alternate<stm32f4xx_hal::gpio::AF7>>, stm32f4xx_hal::gpio::gpioa::PA3<stm32f4xx_hal::gpio::Alternate<stm32f4xx_hal::gpio::AF7>>)>`
```

therealprof:
> > If I'm not mistaken this HAL impl should allow you to `writeln!` directly onto the `serial`. Splitting should only be necessary if you want to read and write in different parts of your application.
> 
> I guess my n00bness strikes again: I saw that in other examples and assumed you actually had to split. But then again, commenting the `split` line and replacing `tx` with `serial` (I guess you meant that) will not compile:
> 
> ```rust
> writeln!(serial, "value: {}\r", value).unwrap();
>    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `stm32f4xx_hal::serial::Serial<stm32f4xx_hal::stm32::USART2, (stm32f4xx_hal::gpio::gpioa::PA2<stm32f4xx_hal::gpio::Alternate<stm32f4xx_hal::gpio::AF7>>, stm32f4xx_hal::gpio::gpioa::PA3<stm32f4xx_hal::gpio::Alternate<stm32f4xx_hal::gpio::AF7>>)>`
> ```

Weird. No, not your fault, the implementation is missing. :(

nebelgrau77:
> Why not show receiving data as well? That's usually the more tricky part...

Mainly because I don't know, how :) I just wanted to add a simple example that I use a lot, and that was missing. For other beginners :)

therealprof:
If you could run `cargo fmt` and squash the commits, we can merge this.

therealprof:
> Is it to satisfy clippy, because otherwise it will suggest to remove that entirely?

Yes, the compiler (and hence clippy) cannot see that anything from that crate is actually used (because it contains a panic handler which will only be added very late in the compilation process) so it is important to `use` it but by renaming it to `_` the compiler (and clippy) won't whine about it.

nebelgrau77:
> If you could run `cargo fmt` and squash the commits, we can merge this.

I'd love to but I'm not sure, how: I see "Merging is blocked: Merging can be performed with 1 approving review" - unless it has nothing to do with that?

therealprof:
In a nutshell I'd ask you to:
- Run `cargo fmt`
- Commit the resulting change
- Do a `git reset --soft origin/master`
- Do a `git commit -a`
- Do a git push --force onto your PR branch

therealprof:
@nebelgrau77 Close. 😅 The history was messed up so to be on the safe side I rebased your branch.

therealprof (approved):

therealprof:
Thanks for sticking with us.

nebelgrau77:
> @nebelgrau77 Close.  The history was messed up so to be on the safe side I rebased your branch.

Thanks! I knew it was to good to be true :D

PR #215, opened 2020-09-20
Add fmt::Write impl for Serial<USART, PINS>

therealprof:
This allows to send formatted output directly to the serial port instead
of having to split it as found in #214.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -1862,12 +1862,26 @@ halUsart! {
     UART10: (uart10, apb2enr, 7, uart10en, pclk2),
 }
 
+impl<USART, PINS> fmt::Write for Serial<USART, PINS>
+where
+    Serial<USART, PINS>: serial::Write<u8>,
+{
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        s.as_bytes()
+            .iter()
+            .try_for_each(|c| block!(self.write(*c)))
+            .map_err(|_| fmt::Error)
+    }
+}
+
 impl<USART> fmt::Write for Tx<USART>
 where
     Tx<USART>: serial::Write<u8>,
 {
     fn write_str(&mut self, s: &str) -> fmt::Result {
-        let _ = s.as_bytes().iter().map(|c| block!(self.write(*c))).last();
-        Ok(())
+        s.as_bytes()
+            .iter()
+            .try_for_each(|c| block!(self.write(*c)))
+            .map_err(|_| fmt::Error)
     }
 }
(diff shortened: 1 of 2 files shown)


src/serial.rs
```diff
@@ -1862,6 +1862,16 @@ halUsart! {
     UART10: (uart10, apb2enr, 7, uart10en, pclk2),
 }
 
+impl<USART, PINS> fmt::Write for Serial<USART, PINS>
+where
+    Serial<USART, PINS>: serial::Write<u8>,
+{
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        let _ = s.as_bytes().iter().map(|c| block!(self.write(*c))).last();
```
thalesfragoso, line 1870: Hmm, I would really prefer a `for loop` in here, I have the impression that it's easier to read and more idiomatic in this case, and not sure if the fact that `last()` keeps track of the last valid value (i.e. a not `None` one) incurs a minimal performance hit, although that bother me less than the "easiness to read".

TeXitoi:
```suggestion
        s.as_bytes().iter().for_each(|c| block!(self.write(*c)));
```

But I also prefer a for loop.

therealprof:
I agree that there's a ton of room for improvement but I actually only wanted to address the glaring omission of being able to `Write` directly to the `Serial` without having to split it first. ;)

src/serial.rs
TeXitoi: Well, it didn't typecheck, sorry, block doesn't return ()

therealprof: No worries. How do you like the current version?

src/serial.rs
```diff
@@ -1862,12 +1862,22 @@ halUsart! {
     UART10: (uart10, apb2enr, 7, uart10en, pclk2),
 }
 
+impl<USART, PINS> fmt::Write for Serial<USART, PINS>
+where
+    Serial<USART, PINS>: serial::Write<u8>,
+{
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        let _ = s.as_bytes().iter().try_for_each(|c| block!(self.write(*c)));
+        Ok(())
```
TeXitoi, line 1871: Now I'm afraid it doesn't work, but I suppose it's better if it does:

```suggestion
        ls.as_bytes().iter().try_for_each(|c| block!(self.write(*c)))
```

Now, that's not really critical ;-)

therealprof: I don't think I follow. What do you mean?

TeXitoi: I thought that the error was Infaillible,  but it's not. So the correct implementation is:

```suggestion
        s.as_bytes().iter().try_for_each(|c| block!(self.write(*c))).map_err(|_| fmt::Error)
```

Else, the serial error will be just silently skipped.

therealprof: I am aware. But completely revamping the implementation was not quite the plan...

TeXitoi: yeah, that's why I said it was not really critical, out of subject would be more appropriate.

TeXitoi (approved):

therealprof:
bors r=TeXitoi

PR #219, opened 2020-10-09
Add support for SCSD cards with SDIO

glbsalazar:
Add support for SCSD cards in the SDIO interface
Added the timeout since setting the bus width and frequency was timing out because the card was not ready.

### DIFF
--- a/src/sdio.rs
+++ b/src/sdio.rs
@@ -314,8 +314,15 @@ impl Sdio {
 
     /// Read a block from the card
     pub fn read_block(&mut self, blockaddr: u32, block: &mut [u8; 512]) -> Result<(), Error> {
-        let _card = self.card()?;
+        let card = self.card()?;
 
+        // Always write 1 block of 512 bytes
+        // SDSC cards are byte addressed hence the blockaddress is in multiples of 512 bytes
+        let blockaddr = match card.capacity {
+            CardCapacity::SDSC => blockaddr * 512,
+            _ => blockaddr,
+        };
+        
         self.cmd(cmd::set_block_length(512))?;
         self.start_datapath_transfer(512, 9, true);
         self.cmd(cmd::read_single_block(blockaddr))?;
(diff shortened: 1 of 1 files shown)


jkristell:
Thanks for looking at this. What kind of card is this tested with? Do read and write work?

glbsalazar:
It was tested with a generic unbranded 2GB micro sd card. 
I tested it by writting checking that the CID, CSD, OCR, RCA and SCR made sense and by writing a block and reading it out and checked that the values were correct.

jkristell:
I'm gonna test this with my board just to make sure SDHC cards still works.

But I'm a bit confused, I thought that the `read/write_block` would had to be modified to handle byte addressed cards.

What is the CSD version on the card?

glbsalazar:
It's a V2 card. I also tested it with a SDHC card (8GB) and it was also working, but I can retest everything latter tonight.

What changes were you thinking were needed?

jkristell:
If it's a V2 card then it is block addressed and it explains why reading and writing works, right?

Hmm, maybe this shouldn't be classified as a SDSC card, perhaps we should look at at the csd version too, and then just classify this as a SDHC. There is probably somethings in the spec about this, or this card is just odd :)

glbsalazar:
I might have done something strange, because the spec indeed states it is byte addressed for SDSC (and makes no consideration of version).
I will do some more tests, because looking at them now they did not cover all cases.
I'll try to retest it and update this later today. 
Thanks for the inputs :)

jkristell:
I was a bit surprised a SDSC in microSD format even existed. So let's not rule out this is some strange card from a vendor doing something weird here :).

glbsalazar:
@jkristell I added the proper handle for the SDSC cards, I previously missed some things in the tests.
Thanks for the input :)

jkristell:
This looks like how I imagined the support for SDSC would look like. Thanks for looking into and fixing this!

jkristell:
I don't have write access to the repo, so pinging @therealprof @thalesfragoso and asking them to merge this now.

therealprof:
@jkristell If you're happy, let's get this merged. ;)

jkristell:
I am! :) So let's merge it.

therealprof (approved):
bors r+

PR #220, opened 2020-10-24
Updating ADC to support external reference specifications

### LINKED ISSUE #205: ADC initialization may panic
When constructing ADC3, I observed the initialization function panic. Upon debugging, it appears that the initializer is panicking during `adc::calibrate()` because `vref_samp` is set to zero.

Upon inspection of the STM32F4xx reference manual, I found this snippet:
image

It does not appear that it is valid to read the VREF_INT signal from any ADC other than ADC1.

ryan-summers:
This PR fixes #205 by updating the adc module to not allow access to VREF, VBAT, and the internal temperature sensor from ADC2 and ADC3, since these are only available to ADC1. To accomodate the loss of external reference calibration, the adc configuration has been updated to allow the user to specify the VDDA voltage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Change DMA traits to `embedded-dma`.
 - Use bitbanding during clock enabling and peripheral reset to avoid data races.
 - Add missing `Write` implementation for `Serial` and implemented better error handling.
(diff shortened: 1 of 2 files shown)


therealprof (changes requested):
LGTM but lacks a CHANGELOG entry (as usual). 😅

therealprof (approved):
LGTM, thanks!

bors r+

PR #222, opened 2020-10-27
add blinky_timer_irq example

YruamaLairba:
Hello, i adapted the blinky_timer_irq example of stm32f1xx-hal. 

The source work when in a dedicated project, but i don't know how to make it work "inplace". I mean i can compile it using
 `cargo build --features="rt stm32f411" --example blinky-timer-irq`
 but when i'm using 
`gdb-multiarch -q target/thumbv7em-none-eabihf/debug/examples/blinky-timer-irq`
 and doing 
`target remote :3333`
 i get
 `Dwarf Error: Cannot find DIE at 0xe3b0 referenced from DIE at 0xe7e7 [in module /home/amaury/projet/rust/embed/stm32f4xx-hal/target/thumbv7em-none-eabihf/debug/examples/blinky-timer-irq]`
I don't know what it mean. I tryed to change memory.x and add the build.rs, don't work.


therealprof:
@YruamaLairba The error message you're getting is coming from the debugger, no idea what it doesn't like here. Maybe try replacing `wfi();` by https://doc.rust-lang.org/core/sync/atomic/fn.spin_loop_hint.html since the former sometimes tricks a debugger.

Other than that looks good to me if you could fix the rustfmt error.

YruamaLairba:
for my "DIE" error, it's not related to the `wfi();`. I almost sure it's doesn't comes from the source file since it works when in a seperate project.

therealprof:
> for my "DIE" error, it's not related to the `wfi();`. I almost sure it's doesn't comes from the source file since it works when in a seperate project.

Then the compiler flags might be the culprit. I don't use debuggers, so I can't really help you there I'm afraid.

therealprof (approved):
LGTM, thanks!

N3xed:
@YruamaLairba I've also had these Dwarf errors before and I've fixed them by changing the lto (link-time optimization) to `thin` (or off, at least not `true`) in the dev profile. There something about lto that the debugger doesn't like if the optimization isn't fully on, because it works (at least for me) when using the release profile (where lto is on, and the opt level is high). Hopefully, this helps.

PR #223, opened 2020-11-02
Clear SPI errors on a fresh transmit

jamesmunns:
Previous to this, all errors on the SPI port were latching. Instead, default to clearing errors at the start of every new transaction.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -1070,7 +1070,9 @@ where
 
         if txe {
             // NOTE(write_volatile) see note above
-            unsafe { ptr::write_volatile(&self.spi.dr as *const _ as *mut u8, byte); }
+            unsafe {
+                ptr::write_volatile(&self.spi.dr as *const _ as *mut u8, byte);
+            }
             Ok(())
         } else {
             Err(nb::Error::WouldBlock)
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -1051,21 +1051,30 @@ where
     }
 
     fn send(&mut self, byte: u8) -> nb::Result<(), Error> {
-        let sr = self.spi.sr.read();
+        // Read from DR register to clear OVR
+        let _ = self.spi.dr.read();
 
-        Err(if sr.ovr().bit_is_set() {
-            nb::Error::Other(Error::Overrun)
-        } else if sr.modf().bit_is_set() {
-            nb::Error::Other(Error::ModeFault)
-        } else if sr.crcerr().bit_is_set() {
-            nb::Error::Other(Error::Crc)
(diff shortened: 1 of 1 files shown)


src/spi.rs
```diff
@@ -1051,21 +1051,32 @@ where
     }
 
     fn send(&mut self, byte: u8) -> nb::Result<(), Error> {
-        let sr = self.spi.sr.read();
+        // Read from DR register to clear OVR
+        let _ = self.spi.dr.read();
```
ryan-summers, line 1055: I would feel more comfortable if we only did this read upon overrun detection. I'm curious if there's a potential race condition here of reading the DR right as we finish shifting out the current TX byte (e.g. we have received a byte for the current transfer, but haven't quite finished sending ours?). Not sure if this can really happen, but it wouldn't hurt to be careful.

src/spi.rs
```diff
@@ -1051,21 +1051,32 @@ where
...
+        // Read from SR to clear CRCERR and OVR
+        let mut txe = false;
+        self.spi.sr.modify(|r, w| {
+            txe = r.txe().bit_is_set();
+
+            if r.crcerr().bit_is_set() {
+                w.crcerr().clear_bit();
+            }
+            w
+        });
```
ryan-summers, line 1066: Right now, this just blindly clears any errors and doesn't indicate that they occurred in the first place. We still need to indicate that the errors happened after clearing them - it's up to the user to determine what to do after the error is detected.

ryan-summers (changes requested):
Minor

src/spi.rs
therealprof: ACK. Unfortunately for that to be useful we need proper generic error types which are still stuck in limbo.

jamesmunns:
My plan to change this is to check errors at the END of every send, and report them then, while also clearing errors at the START of every transaction. Let me know if this works for you @ryan-summers and @therealprof

jamesmunns:
Note to self: add to changelog

therealprof:
> My plan to change this is to check errors at the END of every send, and report them then, while also clearing errors at the START of every transaction. Let me know if this works for you @ryan-summers and @therealprof

Yes, that seems sensible. Also cf. https://github.com/stm32-rs/stm32f0xx-hal/pull/131 and https://github.com/stm32-rs/stm32f0xx-hal/issues/130 where we also discussed error handling as a tangent.

therealprof:
> Note to self: add to changelog

Indeed. ;)

jamesmunns:
Superceded by #269. Thanks @riskable!

PR #224, opened 2020-11-02
Correct Serial Baudrate calculation

jamesmunns:
I think the baudrate calculation before was... ~~really wrong~~ slightly misunderstood.

As far as I can tell, this new method is correct according to the stm32F411 datasheet, I'm not sure if the calculation is consistent across all stm32f4xx devices.

This was validated with an stm32f411, a pclk2 of 96MHz, and a baudrate of 12mbps.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -1424,10 +1424,55 @@ macro_rules! halUsartImpl {
                         bb::set(&rcc.$apbXenr, $rcc_bit);
                     }
 
+                    let pclk_freq = clocks.$pclkX().0;
+                    let baud = config.baudrate.0;
+
+                    // The frequency to calculate USARTDIV is this:
+                    //
+                    // (Taken from STM32F411xC/E Reference Manual,
+                    // Section 19.3.4, Equation 1)
+                    //
+                    // 16 bit oversample: OVER8 = 0
+                    // 8 bit oversample:  OVER8 = 1
+                    //
+                    // USARTDIV =          (pclk)
+                    //            ------------------------
+                    //            8 x (2 - OVER8) x (baud)
+                    //
+                    // BUT, the USARTDIV has 4 "fractional" bits, which effectively
+                    // means that we need to "correct" the equation as follows:
+                    //
+                    // USARTDIV =      (pclk) * 16
+                    //            ------------------------
+                    //            8 x (2 - OVER8) x (baud)
+                    //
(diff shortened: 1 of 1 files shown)


src/serial.rs
```diff
@@ -1424,10 +1424,55 @@ macro_rules! halUsartImpl {
...
+                    // right one bit
+
                     // Calculate correct baudrate divisor on the fly
-                    let div = (clocks.$pclkX().0 + config.baudrate.0 / 2)
-                        / config.baudrate.0;
-                    usart.brr.write(|w| unsafe { w.bits(div) });
+                    let (over8, div) = if (pclk_freq / 16) >= baud {
+                        // We have the ability to oversample to 16 bits, take
+                        // advantage of it.
+                        let div = pclk_freq / baud;
```
aurelj, line 1457: Not sure why you removed the rounding that was in the original code ?

eg. with a 48 MHz pclk and a 115200 baudrate:
original code: div=417 actual_baudrate=115108
your code: div=416 actual_baudrate=115385

So with your code the actual baudrate is further away from the target baudrate than with the original code.

src/serial.rs
```diff
@@ -1424,10 +1424,55 @@ macro_rules! halUsartImpl {
...
+                        (false, div)
+                    } else if (pclk_freq / 8) >= baud {
+                        // We are close enough to pclk where we can only
+                        // oversample 8.
+                        let div = (pclk_freq * 2) / baud;
```
aurelj, line 1462: You should do some rounding here too.

aurelj:
> I think the baudrate calculation before was... really wrong.

The baudrate calculation was correct except for very high baudrate (or very low pclk), that's why it wasn't noticed until now.

Anyway, it is indeed nice to add support for baudrate > pclk / 16.

jamesmunns:
Ah! I didn't understand that the `+ baud` term was to better round. That was part of my original confusion.

I'll see if I can add that back to get a better approximation.

jamesmunns:
Restored the rounding logic, and added comments so people won't get confused (like me) in the future. Thanks for the feedback @aurelj!

aurelj:
Great ! Thanks !

aurelj (approved):

therealprof (approved):
Excellent documentation, thanks @jamesmunns.

@aurelj Don't be afraid to merge PRs. ;)

bors r+

PR #225, opened 2020-11-04
Add CRC peripheral

jamesmunns:


Disasm:
I'd prefer to call functions `init` (performs reset in this case) and `update` (feeds without resetting) as they are commonly used in other libraries.

therealprof:
I tend to agree with @Disasm here that `feed` is a bit boutique. 😅

jamesmunns:
Addressed comments!

therealprof (changes requested):
Missing CHANGELOG entry, other than that looks very nice to me.

therealprof (approved):
Thanks a lot!

bors r+

PR #228, opened 2020-11-06
Add rtc support

mdrobnak:
This PR adds RTC support for the stm32f4 series SoCs. This is based on the stm32f3 code by @David-OConnor, and ST reference examples. Set and get of time, and individual hour/minute/second/day/month/year all tested successfully. Have not tested full date set, but uses the same code and should work fine.
Personally tested on a ST Nucleo F446RE board. Very similar to https://github.com/stm32-rs/stm32f7xx-hal/pull/96
Has CHANGELOG entry.
Not sure if this will auto-populate reviewers, but will call out @therealprof as I've seen that name on quite a few sign-offs. :)

Please let me know if anything needs to be modified.
Thanks,
-Matt

### DIFF
--- a/src/rtc.rs
+++ b/src/rtc.rs
@@ -1,0 +1,428 @@
+//! Interface to the real time clock. See STM32F303 reference manual, section 27.
+//! For more details, see
+//! [ST AN4759](https:/www.st.com%2Fresource%2Fen%2Fapplication_note%2Fdm00226326-using-the-hardware-realtime-clock-rtc-and-the-tamper-management-unit-tamp-with-stm32-microcontrollers-stmicroelectronics.pdf&usg=AOvVaw3PzvL2TfYtwS32fw-Uv37h)
+
+use crate::pac::{PWR, RTC};
+use crate::rcc::{APB1, BDCR};
+use core::convert::TryInto;
+use rtcc::{Datelike, Hours, NaiveDate, NaiveDateTime, NaiveTime, Rtcc, Timelike};
+
+/// Invalid input error
+#[derive(Debug)]
+pub enum Error {
+    InvalidInputData,
+}
+
+pub const LSE_BITS: u8 = 0b01;
+
+pub struct Rtc {
+    pub regs: RTC,
+}
+
+impl Rtc {
+    /// Create and enable a new RTC, and configure its clock source and prescalers.
+    /// From AN4759, Table 7, when using the LSE (The only clock source this module
+    /// supports currently), set `prediv_s` to 255, and `prediv_a` to 127 to get a
+    /// calendar clock of 1Hz.
+    /// The `bypass` argument is `true` if you're using an external oscillator that
+    /// doesn't connect to `OSC32_IN`, such as a MEMS resonator.
+    pub fn new(
+        regs: RTC,
+        prediv_s: u16,
+        prediv_a: u8,
+        bypass: bool,
+        apb1: &mut APB1,
+        bdcr: &mut BDCR,
+        pwr: &mut PWR,
+    ) -> Self {
+        let mut result = Self { regs };
+
+        enable_lse(bdcr, bypass);
+        unlock(apb1, pwr);
+        enable(bdcr);
+        result.set_24h_fmt();
+
+        result.regs.prer.modify(|_, w| unsafe {
+            w.prediv_s().bits(prediv_s);
+            w.prediv_a().bits(prediv_a)
+        });
+
+        result
+    }
+
+    /// Sets calendar clock to 24 hr format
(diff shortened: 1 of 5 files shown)


Logiase:
Thanks for adding support for RTC, it really helps me.

src/rtc.rs
```diff
@@ -0,0 +1,458 @@
...
+        Ok(NaiveTime::from_hms(
+            hours.into(),
+            minutes.into(),
+            seconds.into(),
+        ))
+    }
+
+    fn get_weekday(&mut self) -> Result<u8, Self::Error> {
+        let dr = self.regs.dr.read();
+        let weekday = bcd2_decode(dr.wdu().bits(), 0x00);
```
twistedfall, line 330: Shouldn't the order of the arguments be inverse here?
```suggestion
        let weekday = bcd2_decode(0x00, dr.wdu().bits());
```
or maybe even better just:
```suggestion
        let weekday = dr.wdu().bits();
```

mdrobnak: Reviews are definitely useful - I don't think I used the day-of-week functionality at all in my code, and I may have skipped reviewing that it worked correctly. I'll check this out in the next few days.

David-OConnor:
Hey - I'm watching this, and will add any changes you make to my F3 fork.

Of note, I recently added better support for Wakeups and Alarms; ie you enter the time in seconds, and it sets the `wucksel` and `wutr` registers appropriate.

src/rtc.rs
twistedfall: Will you please also check the following: in section 17.3.6 Reading the calendar, there is:
```
To ensure consistency between the 3 values, reading either RTC_SSR or RTC_TR locks
the values in the higher-order calendar shadow registers until RTC_DR is read.
```
I noticed this behavior in my own implementation of RTC, so I had to make sure I read DR after I read SSR or TR, but I can't see this provision in you implementation. So e.g. if I would read only time for several days in a row then a DR read will return a date from when TR was first read and not the current date.

mdrobnak: Yes, this is why I reversed the order in get_datetime from date, then time, to time, then date. Ugh, that really potentially muddies the code quite a bit.

twistedfall: Well, you can just make sure to just read (and discard) DR in every time-only function after the TR is read.

therealprof:
@mdrobnak Any update on this?

mdrobnak:
> @mdrobnak Any update on this?

Once I'm finished getting bxcan integrated correctly, I will come back to this. Definitely this week.

mdrobnak:
I'm double checking this now, but I think this is good at this point. I'll mark it ready for review when I think that is the case.

mdrobnak:
This works. Tested with Weekday set to 7, and date set to 2021-01-10, correctly rolled to 2021-01-11 and weekday of 1.

mdrobnak:
@therealprof What else is needed from me at this point?

src/rtc.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// calendar clock of 1Hz.
+    /// The `bypass` argument is `true` if you're using an external oscillator that
+    /// doesn't connect to `OSC32_IN`, such as a MEMS resonator.
+    pub fn new(
+        regs: RTC,
+        prediv_s: u16,
+        prediv_a: u8,
+        bypass: bool,
+        apb1: &mut APB1,
+        bdcr: &mut BDCR,
```
xoviat, line 35: Taking feedback that I received on my PR, you should not accept the ABP1 and BDCR structs. Rather, you should use bit banding to enable the bits.

xoviat (changes requested):
Some thoughts.

src/rtc.rs
mdrobnak: I have made the requested changes, hopefully this is what was being looked for.

mdrobnak:
Is there anything else needed here?

src/rtc.rs
```diff
@@ -0,0 +1,480 @@
+//! Interface to the real time clock. See STM32F303 reference manual, section 27.
```
therealprof, line 1: Hm... Feels a bit weird to reference the F303 RM here.

mdrobnak: The original code was written for the F3, I didn't modify those comments.

therealprof (approved):
Okay then.

bors r+

therealprof (approved):

mdrobnak:
Awesome! Thanks for merging @therealprof . @David-OConnor if you could open a new PR with the alarm stuff I'm sure that'd be appreciated by many users.

PR #229, opened 2020-11-08
timer: trigger update event to load the registers

andresv:
@thalesfragoso suggested this fix in matrix.

rtic init:
```rust
let mut timer = Timer::tim5(ctx.device.TIM5, 10.hz(), clocks);
timer.listen(Event::TimeOut);
pb5.set_high().ok();
```

rtic isr:
```rust
#[task(binds = TIM5, resources = [led, sm, pb5])]
fn timer5_isr(ctx: timer5_isr::Context) {
    let tim = unsafe { &(*stm32::TIM5::ptr()) };
    ctx.resources.pb5.toggle().ok();
    tim.sr.modify(|_, w| w.uif().clear_bit());
}
```
PB5 pin trace, notice the glitch:
Screenshot 2020-11-08 at 17 14 06

With this fix timings are perfect:
Screenshot 2020-11-08 at 19 35 28

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Stability fixes related to SD card write
+- Fixed issue where timer generated a spurious interrupt after start
 
 ## [v0.8.3] - 2020-06-12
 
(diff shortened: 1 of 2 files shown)


thalesfragoso (commented):
Thanks for finding this and making the PR, could you add an entry to the changelog under the `Fixed` part ?

thalesfragoso (approved):
Thanks again!

bors r+

PR #231, opened 2020-11-18
Implement unsafe overclocking safety overrides

rukai:
I am porting a C firmware that slightly overclocks the following values on an stm32f439.
* SYSCLK 192MHz (max 180MHz)
* APB1/PCLK1/PPRE1_register 48Mhz (max 45MHz)
* APB2/PCLK2/PPRE2_register 96Mhz (max 90MHz)

I need to maintain these clock speeds in the rust port, so my usage will look like this:

```rust
    let clocks = unsafe {
        rcc.cfgr
            .use_hse(4.mhz())
            .sysclk(192.mhz())
            .hclk(192.mhz())
            .pclk1(48.mhz())
            .pclk2(96.mhz())
            .sysclk_allow_overclock()
            .pclk1_allow_overclock()
            .pclk2_allow_overclock()
            .freeze()
        };
```

Would love to hear suggestions for better naming of the builder method names.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -19,9 +19,6 @@ impl RccExt for RCC {
                 pclk2: None,
                 sysclk: None,
                 pll48clk: false,
-                pclk1_overclock: false,
-                pclk2_overclock: false,
-                sysclk_overclock: false,
             },
         }
     }
@@ -127,9 +124,6 @@ pub struct CFGR {
     pclk2: Option<u32>,
     sysclk: Option<u32>,
     pll48clk: bool,
-    pclk1_overclock: bool,
-    pclk2_overclock: bool,
-    sysclk_overclock: bool,
 }
 
 impl CFGR {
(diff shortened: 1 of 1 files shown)


TeXitoi:
I'll do a `unsafe pub freeze_unchecked(self)`

rukai:
Sure, sounds like you are implementing it yourself, so should I close this PR?

TeXitoi:
That's just a naming proposition.

rukai:
A single `freeze_unchecked` method instead of the 3 `*_allow_overclock` methods would be fine by me.

Would it also suppress this assert? `assert!(vco_in >= 1_000_000 && vco_in <= 2_000_000);`
I have no idea if that is a good idea or not, the C code I am porting doesnt break this assert.

I'll wait for a maintainer to say which api they prefer before I rewrite it.

therealprof:
`freeze_unchecked` sounds like a better idea to me, too.

rukai:
done!

TeXitoi (approved):

PR #232, opened 2020-11-20
Add syscfg driver to enforce clock enable 

Disasm:
This should prevent terrible user experience when SYSCFG is used without its clock enabled.

Note that this is a breaking change.


therealprof (changes requested):
Great idea! Thanks.

But continuing on with tradition this is lacking a CHANGELOG.md entry. 😅

Disasm:
@therealprof fixed :) I should probably pull a changelog requirement CI rule from probe-rs.

therealprof:
> @therealprof fixed :) I should probably pull a changelog requirement CI rule from probe-rs.

Such a thing exists? That'd be a great addition to all WG repos...

therealprof (approved):
Thanks a lot!

bors r+

Disasm:
https://github.com/probe-rs/probe-rs/pull/428/files
However, it uses `secrets.GITHUB_TOKEN`, I'm not sure why. Probably needs additional setup.

Issue #226, opened 2020-11-05
SPI DMA transfer size support


albru123:
With current DMA implementation, it's impossible to configure DMA with 16-bit transfer size for SPI. It's also impossible to implement a wrapper type around SPI to define the u16 size manually, as you cannot implement the necessary [`DMASet`](https://github.com/stm32-rs/stm32f4xx-hal/blob/385cb9707994d41a6199d1d139b49cfa4490a87d/src/dma/traits.rs#L313) outside of the crate.

With the current structure, it would be necessary for [`PeriAddress`](https://github.com/stm32-rs/stm32f4xx-hal/blob/385cb9707994d41a6199d1d139b49cfa4490a87d/src/dma/traits.rs#L187) to accept transfer size as a generic type argument instead of having it as an associated type to enable multiple possible implementations for SPI (as you cannot implement `PeriAddress` twice with different `MemSize` when `MemSize` is an associated type). This change then propagates upwards up to the [`Transfer<>`](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/dma/mod.rs#L846) type where another generic type argument defining the transfer size would be necessary.

This in fact makes perfect sense, as you cannot always assume the transfer size from the peripheral upon which you want to perform the DMA transfers.

```rust
type DmaTransfer = Transfer<Stream4<DMA1>, Channel0, SPI2, MemoryToPeripheral, &'static mut [u16; 64], u16 /* transfer size */>;
```

This is how the new `Transfer<>` type would probably look like when implemented naively like I proposed. There was also a proposition to hold the DMA transfer size associated with the SPI peripheral (as it usually depends on the SPI frame size), but I think it's not always desirable, as you could also use a [u8] source (something like BBQueue) to feed SPI with u16 frame size.

I think the transfer size parameter should be made a bit less confusing at least, by using something like an enum to define the desired size, if implemented like this.

thalesfragoso:
Thanks for the report and investigation.

Having the `MemSize` as an associated type was probably an oversight on my part, I think it was because we already needed new types over some peripherals from day one (for timers with multiple possible registers), then I thought making a new type for configuration would make it clearer, but I didn't notice the problem with `DmaSet` and tuples.

So, I agree that having `MemSize` as a generic parameter is probably the best solution here.

> I think the transfer size parameter should be made a bit less confusing at least, by using something like an enum to define the desired size, if implemented like this.

[...]

PR #237, opened 2020-11-23
DMA: Fix problems with DMASet and timer wrappers

### LINKED ISSUE #227: TIM DMA transfer support
Hi,
I'm trying to implement the Dshot protocol, similar to the one WS2812B uses.
This is the code where I already fail:
```rust
let dma1_streams = dma::StreamsTuple::new(device.DMA1);
let dma_transfer = dma::Transfer::init(
    dma1_streams.4,
    device.TIM3.ccr1,
    cortex_m::singleton!(: [u16; 18] = [0; 18]).unwrap(),
    None,
    dma::config::DmaConfig::default()
        .fifo_enable(true)
        .memory_increment(true),
);
```
Unfortunately it doesn't let me compile and returns an error message:
```
the trait bound `stm32f4::generic::Reg<u32, stm32f4xx_hal::stm32::tim3::_CCR>: stm32f4xx_hal::dma::traits::PeriAddress` is not satisfied
the trait `stm32f4xx_hal::dma::traits::PeriAddress` is not implemented for `stm32f4::generic::Reg<u32, stm32f4xx_hal::stm32::tim3::_CCR>`
```
I've seen in the source code that it requires the struct `CCR1<pac::TIM3>` to use `TIM3_CH1`, but where do I get it from?

thalesfragoso:
The initial idea of the DMASet trait was that people could __unsafely__ implement it themselves from outside of the HAL. Unfortunately, we missed the problem with the fact that tuples are considered a foreign type. This PR drops the tuple approach but still maintains the check that DMASet provides.

I also added a fix for #227.
Fixes #227 

This is also a possible solution to #226, but it takes a different approach, it still forces people (or the HAL) to make a new type if they want a "non-default" memory size (or address). This is more verbose than just dropping the associated type of `PeriAddress` in favor of a generic parameter, but, IMO, the verbosity can be an advantage here.

Marking as a draft because I still want to make sure this works as intended and that there isn't another corner case, I also would like the opinion of @albru123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,8 +39,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added support for hardware-based CRC32 functionality
 
 ### Fixed
+
 - Stability fixes related to SD card write
 - Fixed issue where timer generated a spurious interrupt after start
+- Allow implementations for DMASet from outside the crate
+- DMA: Make it possible to create the wrapper types for the timers
 
 ## [v0.8.3] - 2020-06-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Stability fixes related to SD card write
 - Fixed issue where timer generated a spurious interrupt after start
-- Allow implementations for DMASet from outside the crate
-- DMA: Make it possible to create the wrapper types for the timers
+- Allow implementations for DMASet from outside the crate [#237]
+- DMA: Make it possible to create the wrapper types for the timers [#237]
+- DMA: Fix some `compiler_fences` [#237]
+
+[#237]: https://github.com/stm32-rs/stm32f4xx-hal/pull/237
 
 ## [v0.8.3] - 2020-06-12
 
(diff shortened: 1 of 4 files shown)


almusil:
Hello,
any plan when this might be finished? I am having hard time to have DMA for ADC that actually does not own the ADC so I can configure ADC through the hal.

thalesfragoso:
> I am having hard time to have DMA for ADC that actually does not own the ADC so I can configure ADC through the hal.

I don't think this PR will solve this. For that you need an implementation on the HAL abstraction itself (and maybe some new/modified methods, etc), similar to what was done for the serial. Each HAL peripheral abstraction will need to support DMA by itself.

almusil:
Of course it would be better to have direct integration with HAL Adc, but with this PR I am able to at least have workaround using something like this:
```
pub(crate) struct AdcAddress;

unsafe impl PeriAddress for AdcAddress {
    type MemSize = u16;

    fn address(&self) -> u32 {
        unsafe { &(*ADC1::ptr()).dr as *const _ as u32 }
    }
}

unsafe impl DMASet<Stream0<DMA2>, Channel0, PeripheralToMemory> for AdcAddress {}
```

thalesfragoso:
I think this is ready for review, @albru123 have you had the chance to try this to see if it solves your issue ?

thalesfragoso:
Rebased.

albru123:
I'm checking it right now, but feel free to push this forward (I'm a bit busy so if something goes wrong I might not be able to respond in time).

therealprof (approved):
LGTM, Let's move this forward.

bors r+

PR #238, opened 2020-11-30
Port MonoTimer from stm32f1xx-hal

bschwind:
This is a straight-up copy of the `MonoTimer` from the `stm32f1xx-hal` crate:

https://github.com/stm32-rs/stm32f1xx-hal/blob/d976ba2312fae2f1b15707573bc5d82ea00ba73f/src/time.rs#L229-L278

I wanted the same functionality in a stm32f4-based project. I tested it on an `stm32f411CE` [WeAct USB C board](https://stm32-base.org/boards/STM32F401CEU6-WeAct-Black-Pill-V3.0), and it seems to behave properly.


therealprof (changes requested):
Missing CHANGELOG entry. Otherwise LGTM.

bschwind:
@therealprof Thanks for the review, I've updated the `CHANGELOG`.

therealprof (dismissed):
LGTM, thanks.

bors r+

therealprof (approved):
Let's try again...

bors r+

PR #239, opened 2020-12-10
Implement flash read/erase/program

astro:


thalesfragoso (changes requested):
Thanks for the PR, nice work. I left a few comments, but something that I didn't quite understand from the manual is this:

> It is not allowed to program data to the Flash memory that would cross the 128-bit row
boundary. In such a case, the write operation is not performed and a program alignment
error flag (PGAERR) is set in the FLASH_SR register.

This seems to bound the programming operation to a maximum of ~~64~~ 16 bytes (and all inside the same 128-bit aligned "row"), I see that you are checking for this error, but I think this is also important to document.

And also, missing changelog entry.

src/flash.rs
```diff
@@ -0,0 +1,152 @@
...
+            w
+                .psize().bits(PSIZE_X8)
+                // no sectore erase
+                .ser().clear_bit()
+                // programming
+                .pg().set_bit()
+        });
+        for (i, byte) in bytes.into_iter().enumerate() {
+            unsafe {
+                *ptr.wrapping_add(i) = byte;
```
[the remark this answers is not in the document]
astro: Right, thanks!

astro:
I have now taken care of the 128-bit row alignment.

TheRealBluesun:
Really well-structured, nice work!  This makes me want to migrate your work to the `stm32f1xx-hal` crate as well :smile:.

Were you working on read/write protection?  If not, I'll gladly take that up. I just "finished" a [basic implementation](https://github.com/stm32-rs/stm32f1xx-hal/pull/299) for the stm32f1xx-hal, I imagine the F4 is very similar.

Hope to see this merged soon!

src/flash.rs
```diff
@@ -0,0 +1,165 @@
...
+        });
+        self.wait_ready();
+        self.ok()
+    }
+
+    /// Program bytes with offset into flash memory,
+    /// aligned to 128-bit rows
+    pub fn program<I>(&mut self, mut offset: usize, mut bytes: I) -> Result<(), Error>
+    where
+        I: Iterator<Item = u8>,
```
TheRealBluesun, line 108: I think this may be more easily used as `IntoIterator`.

src/flash.rs
```diff
@@ -0,0 +1,165 @@
...
+        unsafe { slice::from_raw_parts(ptr, self.len()) }
+    }
+    /// Unlock flash for erasing/programming until this method's
+    /// result is dropped
+    fn unlocked(&mut self) -> UnlockedFlash;
+}
+
+impl FlashExt for FLASH {
+    fn address(&self) -> usize {
+        0x0800_0000
```
TheRealBluesun, line 52: Could this be a static fn instead? Same question applies to `len()`.

therealprof:
@astro Thanks for the PR. I like the ideas proposed by @TheRealBluesun, what are your opinions on that? Also the unsafe warnings are a bit curious; I haven't checked whether they're actually valid or caused by differences in the patched SVD files.

src/flash.rs
thalesfragoso: `read` needs to borrow `self`, otherwise the user would be able to have a `&` and `&mut` to the same memory.

src/flash.rs
thalesfragoso: Isn't `Iterator` more flexible than `IntoInterator` ?

src/flash.rs
thalesfragoso: Sorry, not `&mut`, but they would be able to mutate the memory while it has a live immutable reference to it.

src/flash.rs
therealprof: No. `IntoIterator` allows you to work with collection types as well as `Iterator` (which automatically implements `IntoIterator`) which is super-handy (but probably not that important in this particular usecase).

thalesfragoso: Oh, just realized that the suggestion was with `Item = &'a u8`

src/flash.rs
TheRealBluesun: > read needs to borrow self, otherwise the user would be able to have a & and &mut to the same memory.
> Sorry, not `&mut`, but they would be able to mutate the memory while it has a live immutable reference to it.

I don't think you're carrying around a reference to memory, though; only the results of a read (i.e. copied bytes). I believe I see what you mean if you think of the entirety of Flash as an object to which you have a slice into, but I don't think very much is gained from treating it that way. I'm really curious to hear what you/others think about it.  I do see how this could become a problem in environments where two threads could simultaneously access flash, though.

src/flash.rs
TheRealBluesun: It may be best to just pass a `&[u8]`, even. Ultimately, I don't think you gain much by using `IntoIter<Item = &u8>` really. Definitely willing to be convinced (insofar as my opinion matters, hah!).

src/flash.rs
thalesfragoso: > I see what you mean if you think of the entirety of Flash as an object to which you have a slice into, but I don't think very much is gained from treating it that way.

`read` is doing exactly that, it gives the user a `&[u8]` pointing directly to flash memory. Without the borrow, the user would be able to create an `UnlockedFlash` and modify the flash contents, and then use the same `&[u8]` after, which is UB in rust, since you can't mutable alias with references.

thalesfragoso: `read` doesn't copy. But even if it did (which isn't really desirable), without borrowing `Self` you couldn't prevent the user from using `read` on main and `program` in an interrupt, which is a data race.

TheRealBluesun: You are right -- I was thinking of my own implementation for `read()`, not the one included here :man_facepalming: .  Apologies.

xoviat:
@therealprof Is there an update on this?

therealprof:
> @therealprof Is there an update on this?

@thalesfragoso needs to withdraw the concern, otherwise bors will not let me merge it.

burrbull:
What status of this?

samcrow:
I have no authority here, but I've been using a forked version with these changes for a while with no problems. I'd love to see this merged so I can switch back to the main (non-forked) source.

Can I do anything to help with this? Do we need someone to create a new pull request with the same changes that won't be stuck?

burrbull:
The code looks good enough for me.
I can merge this after rebase and fixing warnings (inc. clippy) if they are present.

src/flash.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /// Program bytes with offset into flash memory,
+    /// aligned to 128-bit rows
+    pub fn program<I>(&mut self, mut offset: usize, mut bytes: I) -> Result<(), Error>
+    where
+        I: Iterator<Item = u8>,
+    {
+        let ptr = self.flash.address() as *mut u8;
+        let mut bytes_written = 1;
```
thalesfragoso, line 111: What is the use of `bytes_written` ?

thalesfragoso:
We probably should change the `IntoInterator` thing to use `Item = &u8`, update the branch and get rid of the warnings, I suggest getting in touch with the author and/or creating a new PR on top of this one.

PR #240, opened 2020-12-14
I2s clock

YruamaLairba:
Hello, i started to work on setting up i2s clock through the HAL. This is the first step for an i2s HAL implementation

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -119,6 +119,50 @@ pub const PCLK2_MAX: u32 = SYSCLK_MAX / 2;
 /// Maximum APB1 peripheral clock frequency
 pub const PCLK1_MAX: u32 = PCLK2_MAX / 2;
 
+/// Maximum I2S clock frequency
+// TODO check actual value across reference manual
+pub const I2SCLK_MAX: u32 = 192_000_000;
+
+/// Minimum division factor for PLLI2S ouptut clock
+// TODO check actual value across reference manual
+pub const PLLI2S_R_MIN: u8 = 2;
+
+/// Maximum division factor for PLLI2S output clock
+// TODO check actual value across reference manual
+pub const PLLI2S_R_MAX: u8 = 7;
+
+/// Minimum PLLI2S multiplication factor for PLLI2S vco
+// TODO check actual value across reference manual
+pub const PLLI2S_N_MIN: u16 = 50;
+
+/// Maximum PLLI2S multiplication factor for PLLI2S vco
+// TODO check actual value across reference manual
+pub const PLLI2S_N_MAX: u16 = 432;
+
+/// Minimum frequency at PLLI2S vco output
+// TODO check actual value across datasheets
+pub const PLLI2S_N_FREQ_MIN: u32 = 100_000_000;
+
+/// Maximum frequency at PLLI2S vco output
+// TODO check actual value across datasheets
+pub const PLLI2S_N_FREQ_MAX: u32 = 432_000_000;
+
+/// Minimum division factor for PLLI2S input clock
+// TODO check actual value across reference manual
+pub const PLLI2S_M_MIN: u8 = 2;
+
+/// Maximum division factor for PLLI2S input clock
+// TODO check actual value across reference manual
+pub const PLLI2S_M_MAX: u8 = 63;
+
+/// Minimum frequency for PLLI2S input clock
+// TODO check actual value across datasheets
+pub const PLLI2S_M_FREQ_MIN: u32 = 950_000;
+
+/// Maximum frequency for PLLI2S input clock
+// TODO check actual value across datasheets
+pub const PLLI2S_M_FREQ_MAX: u32 = 2_100_000;
+
 struct PLLI2SCFGR {
     r: u8,
     n: u16,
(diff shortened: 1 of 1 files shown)


mgottschlag:
I think the clock configuration interface should not specify R/N/M directly, but should rather specify a target plli2sclk frequency. Most STM32F4 models do not support the PLLI2SM divider but rather use PLLM instead, in which case only specifying PLLI2SR and PLLI2SN would result in varying resulting frequencies depending on the PLLM automatically selected by pll_setup().

Whatever external script the user uses to determine the best clock configuration (w.r.t. your comment in the other pull request) can just output that intermediate plli2sclk frequency.

mgottschlag:
Note: I am currently trying to write HAL wrapper for SAI, so I need basically the same clock configuration code as you do.

Which MCU model are you using? I am trying to figure out the differences in clock configuration - so far I found the following:

- Some models have a separate PLL for SAI, some do not.
- Some have separate M factors for the different PLLs, some do not.
- Some select the I2S clock source in CFGR, some in DCKCFGR.

| Models                      | 405/407/415/417    | 411                | 413/423            | 427/429/437/439    | 446                | 469/479            |
| --------------------------- | ------------------ | ------------------ | ------------------ | ------------------ | ------------------ | ------------------ |
| I2S                         | 2                  | 5                  | 5                  | 2                  | 3                  | 2                  |
| SAI                         | 0                  | 0                  | 1                  | 1                  | 2                  | 1                  |
| Separate PLL for SAI        |                    |                    | :x:                | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| PLLI2SM in RCC_PLLI2SCFGR   | :x:                | :heavy_check_mark: | :heavy_check_mark: | :x:                | :heavy_check_mark: | :x:                |
| PLLSAIM in RCC_PLLSAICFGR   |                    |                    |                    | :x:                | :heavy_check_mark: | :x:                |
| PLLI2SSRC in RCC_PLLI2SCFGR | :x:                | :x:                | :heavy_check_mark: | :x:                | :x:                | :x:                |
| I2SSRC in RCC_CFGR          | :heavy_check_mark: | :heavy_check_mark: | :x:                | :heavy_check_mark: | :x:                | :heavy_check_mark: |
| I2S1/2SRC in RCC_DCKCFGR    | :x:                | :x:                | :heavy_check_mark: | :x:                | :heavy_check_mark: | :x:                |
| SAI1A/BSRC in RCC_DCKCFGR   |                    |                    | :heavy_check_mark: | :heavy_check_mark: | :x:                | :heavy_check_mark: |
| SAI1/2SRC in RCC_DCKCFGR    |                    |                    | :x:                | :x:                | :heavy_check_mark: | :x:                |

Implementing a universal solution might be more difficult than I expected.

mgottschlag:
STM32F410 do not even have an I2S PLL. This is the full table generated from all reference manuals I could find:

| Models                      | 401                | 405/407/415/417    | 410                     | 411                | 412                | 413/423            | 427/429/437/439    | 446                | 469/479            |
| --------------------------- | ------------------ | ------------------ | ----------------------- | ------------------ | ------------------ | ------------------ | ------------------ | ------------------ | ------------------ |
| I2S                         | 2                  | 2                  | 3                       | 5                  | 5                  | 5                  | 2                  | 3                  | 2                  |
| SAI                         | 0                  | 0                  | 0                       | 0                  | 0                  | 1                  | 1                  | 2                  | 1                  |
| I2SPLL                      | :heavy_check_mark: | :heavy_check_mark: | :x:                     | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| SAIPLL                      |                    |                    |                         |                    |                    | :x:                | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| PLLI2SM in RCC_PLLI2SCFGR   | :x:                | :x:                |                         | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :x:                | :heavy_check_mark: | :x:                |
| PLLSAIM in RCC_PLLSAICFGR   |                    |                    |                         |                    |                    |                    | :x:                | :heavy_check_mark: | :x:                |
| PLLI2SSRC in RCC_PLLI2SCFGR | :x:                | :x:                |                         | :x:                | :heavy_check_mark: | :heavy_check_mark: | :x:                | :x:                | :x:                |
| I2SSRC in RCC_CFGR          | :heavy_check_mark: | :heavy_check_mark: | :x:                     | :heavy_check_mark: | :x:                | :x:                | :heavy_check_mark: | :x:                | :heavy_check_mark: |
| I2S1/2SRC in RCC_DCKCFGR    | :x:                | :x:                | :heavy_check_mark: (1x) | :x:                | :heavy_check_mark: | :heavy_check_mark: | :x:                | :heavy_check_mark: | :x:                |
| SAI1A/BSRC in RCC_DCKCFGR   |                    |                    |                         |                    |                    | :heavy_check_mark: | :heavy_check_mark: | :x:                | :heavy_check_mark: |
| SAI1/2SRC in RCC_DCKCFGR    |                    |                    |                         |                    |                    | :x:                | :x:                | :heavy_check_mark: | :x:                |

YruamaLairba:
sorry, i initially open this draft without knowing there was so many difference between f4 chips, and without knowing there was already [another work on i2s](212). I didn't drop it just because the other work don't address yet the i2s clock. As you saw, it's look like a nightmare, so i gave up.

> I think the clock configuration interface should not specify R/N/M directly, but should rather specify a target plli2sclk frequency.

IMHO, specify a target plli2sclk is only useful to keep coherence with the rest on the clock system. I don't find it convenient in practice for many reason:
 - it's complicated for HAL developer to calculate prescalers values
 - It doesn't really simplify the life of HAL users, they have to do some complex calculation to use the desired sampling frequency anyway.
 - it tend to hide the fact you have a difference between specified and real clock while this difference can be relevant.
 
> Which MCU model are you using?

stm32f411 on a nucleo board

> Whatever external script the user uses to determine the best clock configuration (w.r.t. your comment in the other pull request) can just output that intermediate plli2sclk frequency.

I don't understand what you are trying to say here. I actually did those scripts, and it determine bests combination of plli2sclk and i2s configuration to obtain a particular sampling frequency. Not smart though, it just try all combination

mgottschlag:
> IMHO, specify a target plli2sclk is only useful to keep coherence with the rest on the clock system. I don't find it convenient in practice for many reason:

Yeah, you are right. I'll give it a try, it should not actually be that difficult to implement.

mgottschlag:
https://github.com/mgottschlag/stm32f4xx-hal/tree/wip contains a full (although currently completely untested) implementation for generation of I2S and SAI clocks for all MCUs.

The situation is bad, but not quite as bad as I expected at least if we limit ourselves to one I2S and one SAI frequency and do not try to generate more than two audio frequencies from two PLLs. If collecting the PLL target frequencies/requirements is detached from the actual PLL configuration, we have four different "PLL styles" and only few "clock selection styles":

a) We have MCUs without I2S PLL (F410), MCUs with multiple PLLs but shared "M", MCUs where the I2SPLL is used for SAI (F413/423), and MCUs with multiple PLLs and individual "M" dividers.
b) We have MCUs with one or two I2S clocks.

Computational complexity stays as it was if I2S and SAI are not used. It is increased due to the brute-force approach of guessing good PLL configurations if either multiple clocks are generated from the same PLL (F410 if I2S is used, where the code tries up to 60 * 400 PLL configurations, although usually a lot less) and if the SAI divider is automatically determined (in which case up to 32 * 60 configurations are tested - this code path can be optimized quite a bit still). I still have to try how well the code is optimized. The complexity prevents implementation as a const fn, though - that was my first approach.

Note that an additional difference between the MCUs is that the STM32F413/423 have a different SAI clock divider range ([1, 31] instead of [1, 32]).

My next steps are to perform some refactoring and to ensure that everything works on my STM32F429.

YruamaLairba:
@mgottschlag you see very motivated :) . Why you don't create a draft pull request ? This will signal to others that you are working on something. Adding to this, you  can see if you code compile for all target each time you push your work on your github account.

For guessing PLL configurations, it may worth to look at the dark art of procedural macro. You can write one to ensure the guessing job is done at compile time and throwing compile error if a valid configuration cannot be found.

YruamaLairba:
I2s clock configuration is addressed in PR #247

PR #243, opened 2020-12-29
add public is_done method to mirror stm32f1xx-hal

xoviat:
See https://github.com/stm32-rs/stm32f1xx-hal/blob/68fd3d6f282173816fd3181e795988d314cb17d0/src/dma.rs#L302-L304

### DIFF
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -1006,9 +1006,9 @@ where
         f(&mut self.peripheral);
         self.stream.disable()
     }
-    
+
     pub fn is_done(&self) -> bool {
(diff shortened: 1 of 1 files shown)


thalesfragoso:
The users can already use the trait method from anywhere and at anytime without even needing `&self`. This seems unnecessary, unless I'm missing something.

xoviat:
What is the existing code to check that the DMA transfer is done? Is it checking the registers? `get_transfer_complete_flag` is not a public method.

xoviat:
I figured it out.

PR #246, opened 2021-01-02
add can support

xoviat:
This adds support for can with the bxcan crate. This differs slightly from the stm32f1 pull request and the stm32f7 pull request: the `Can::new` method takes ownership of the pins, which requires the user to configure the pins correctly, and doesn't use the afio functionality, which doesn't appear to be present here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,10 +22,11 @@ repository = "https://github.com/stm32-rs/stm32f4xx-hal"
 version = "0.8.3"
 
 [package.metadata.docs.rs]
-features = ["stm32f429", "rt", "usb_fs"]
+features = ["stm32f429", "rt", "usb_fs", "can"]
 targets = ["thumbv7em-none-eabihf"]
 
 [dependencies]
+bxcan = { version = "0.3.0", optional = true }
 cortex-m = ">=0.5.8,<0.7"
 cortex-m-rt = "0.6.10"
 nb = "0.1.2"
@@ -88,6 +89,8 @@ stm32f479 = ["stm32f4/stm32f469", "device-selected"]
 usb_fs = ["synopsys-usb-otg", "synopsys-usb-otg/fs"]
 usb_hs = ["synopsys-usb-otg", "synopsys-usb-otg/hs"]
 
+can = ["bxcan"]
+
 sdio = ["sdio-host"]
 
 [profile.dev]
@@ -138,3 +141,7 @@ required-features = ["rt", "stm32f407"]
 [[example]]
 name = "qei"
 required-features = ["rt", "stm32f411"]
+
+[[example]]
+name = "can-send"
+required-features = ["can", "stm32f405"]
--- a/examples/can-send.rs
+++ b/examples/can-send.rs
@@ -1,0 +1,100 @@
+//! Simple CAN example.
+//! Requires a transceiver connected to PB8, 9 (CAN1) or PB5 PB6 (CAN2).
+
+#![no_main]
+#![no_std]
+
+use panic_halt as _;
+
+use bxcan::filter::Mask32;
+use bxcan::{Frame, StandardId};
+use cortex_m_rt::entry;
+use nb::block;
+use stm32f4xx_hal::{can::Can, pac, prelude::*};
+
+#[entry]
+fn main() -> ! {
+    let dp = pac::Peripherals::take().unwrap();
+
+    let rcc = dp.RCC.constrain();
+
+    // To meet CAN clock accuracy requirements an external crystal or ceramic
+    // resonator must be used. The blue pill has a 8MHz external crystal.
+    // Other boards might have a crystal with another frequency or none at all.
+    rcc.cfgr.use_hse(8.mhz()).freeze();
+
+    let gpiob = dp.GPIOB.split();
+    let mut can1 = {
+        let rx = gpiob.pb8.into_alternate_af9();
+        let tx = gpiob.pb9.into_alternate_af9();
+
+        let can = Can::new(dp.CAN1, (tx, rx));
+
+        bxcan::Can::new(can)
+    };
+    can1.configure(|config| {
+        // APB1 (PCLK1): 8MHz, Bit rate: 500kBit/s, Sample Point 87.5%
+        // Value was calculated with http://www.bittiming.can-wiki.info/
+        config.set_bit_timing(0x001c_0000);
+    });
+
+    // Configure filters so that can frames can be received.
+    let mut filters = can1.modify_filters();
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -26,6 +26,7 @@ features = ["stm32f429", "rt", "usb_fs"]
 targets = ["thumbv7em-none-eabihf"]
 
 [dependencies]
+bxcan = ">=0.3.0"
```
mgottschlag, line 29: bxcan = { version = "0.3.0", optional = true }

and then below "has-can = ["bxcan"]" to make bxcan optional?

mgottschlag:
While the RCC interface looks completely unlike everything else, I'd argue that it is less racy than the existing implementations which just do a read-modify-write cycle in a potentially multitasked (i.e., interrupts of different priority) application.

src/can.rs
```diff
@@ -0,0 +1,125 @@
...
+//! |----------|-------|
+//! | TX       | PB9   |
+//! | RX       | PB8   |
+//!
+//! ### CAN2
+//!
+//! | Function | Pins  |
+//! |----------|-------|
+//! | TX       | PB13  |
+//! | RX       | PB12  |
```
timokroeger, line 19: Those tables do not fully reflect the pins defined on L59.

src/can.rs
```diff
@@ -0,0 +1,125 @@
...
+        $(
+            impl sealed::Sealed for ($tx<Alternate<AF9>>, $rx<Alternate<AF9>>) {}
+            impl Pins for ($tx<Alternate<AF9>>, $rx<Alternate<AF9>>) {
+                type Instance = $PER;
+            }
+        )+
+    }
+}
+
+pins! {
```
timokroeger, line 59: Quick look at the stm32f407 datasheet shows that the can signals can be mapped to even more pins (for larger packages).

src/can.rs
```diff
@@ -0,0 +1,125 @@
...
+pub struct Can<Instance> {
+    _peripheral: Instance,
+}
+
+impl<Instance> Can<Instance>
+where
+    Instance: Enable<Bus = APB1>,
+{
+    /// Creates a CAN interaface.
+    pub fn new<P>(can: Instance, _pins: P, apb: &mut APB1) -> Can<Instance>
```
timokroeger, line 104: Its a common pattern to take ownership of the pins.
For CAN it might be a bit too restrictive though: https://github.com/stm32-rs/stm32f1xx-hal/pull/215 (section *Separate assign_pins()*)

Cargo.toml
```diff
@@ -64,12 +66,13 @@ micromath = "1.0.0"
...
 device-selected = []
+has-can = ["bxcan"]
 rt = ["stm32f4/rt"]
 # Note: stm32f4 has only one feature for some very similar device families,
 # so it's intended for e.g. stm32f405/415 to both enable stm32f4/stm32f405.
 stm32f401 = ["stm32f4/stm32f401", "device-selected"]
-stm32f405 = ["stm32f4/stm32f405", "device-selected"]
-stm32f407 = ["stm32f4/stm32f407", "device-selected"]
+stm32f405 = ["stm32f4/stm32f405", "device-selected", "has-can"]
+stm32f407 = ["stm32f4/stm32f407", "device-selected", "has-can"]
```
thalesfragoso, line 75: Maybe it should default to off, like usb ? Where the user must manually enable the feature in order to use the abstraction.

Cargo.toml
```diff
@@ -26,10 +26,12 @@ features = ["stm32f429", "rt", "usb_fs"]
 targets = ["thumbv7em-none-eabihf"]
 
 [dependencies]
+bxcan = { version = ">=0.3.0", optional = true }
```
thalesfragoso, line 29: Why not just `= 0.3.0` ?

examples/can-send.rs
```diff
@@ -0,0 +1,99 @@
...
+        // APB1 (PCLK1): 8MHz, Bit rate: 500kBit/s, Sample Point 87.5%
+        // Value was calculated with http://www.bittiming.can-wiki.info/
+        config.set_bit_timing(0x001c_0000);
+    });
+
+    // Configure filters so that can frames can be received.
+    let mut filters = can1.modify_filters();
+    filters.enable_bank(0, Mask32::accept_all());
+
+    let can2 = {
```
thalesfragoso, line 45:
```suggestion
    let _can2 = {
```

examples/can-send.rs
```diff
@@ -0,0 +1,99 @@
...
+    let id: u16 = 0x500;
+    loop {
+        test[0] = count;
+        test[1] = 1;
+        test[2] = 2;
+        test[3] = 3;
+        test[4] = 4;
+        test[5] = 5;
+        test[6] = 6;
+        test[7] = 7;
```
thalesfragoso, line 90: Move this to outside the loop ?

src/can.rs
```diff
@@ -0,0 +1,145 @@
...
+    where
+        P: Pins<Instance = Instance>,
+    {
+        Instance::enable(apb);
```
thalesfragoso, line 120: I would prefer the use of bit banding here. It might be good to also reset the peripheral here (with bit banding too)

src/rcc/mod.rs
```diff
@@ -229,6 +235,50 @@ pub const PCLK2_MAX: u32 = SYSCLK_MAX / 2;
...
+    _0: (),
+}
+
+impl APB1 {
+    pub(crate) fn enr(&mut self) -> &rcc::APB1ENR {
+        // NOTE(unsafe) this proxy grants exclusive access to this register
+        unsafe { &(*RCC::ptr()).apb1enr }
+    }
+
+    pub(crate) fn rstr(&mut self) -> &rcc::APB1RSTR {
```
thalesfragoso, line 257: remove ?

thalesfragoso (changes requested):
Thanks for the PR, nice work. I just left a few questions and suggestions, it would be nice if you could also drop the RTC commits.

src/can.rs
```diff
@@ -0,0 +1,135 @@
...
+macro_rules! bus {
+    ($($PER:ident => ($peren:literal),)+) => {
+        $(
+            impl sealed::Sealed for crate::pac::$PER {}
+            impl Enable for crate::pac::$PER {
+                #[inline(always)]
+                fn enable() {
+                    unsafe {
+                        let rcc = &(*RCC::ptr());
+                        bb::set(&rcc.apb2enr, $peren)
```
xoviat, line 86: Whoops!

thalesfragoso (approved):
Thanks again, sorry for the delay.

bors r+

PR #250, opened 2021-01-10
Use CamelCase for gpio::Edge enum variants

Piroro-hs:
ref: [h7xx](https://github.com/stm32-rs/stm32h7xx-hal/pull/179), [l4xx](https://github.com/stm32-rs/stm32l4xx-hal/pull/187), [f1xx](https://github.com/stm32-rs/stm32f1xx-hal/pull/303), [f7xx](https://github.com/stm32-rs/stm32f7xx-hal/pull/111)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `SysCfg` wrapper to enforce clock enable for `SYSCFG`
 - [breaking-change] gpio::ExtiPin now uses `SysCfg` wrapper instead of `SYSCFG`
 - Change `WriteBuffer + 'static` to `StaticWriteBuffer`in the DMA module.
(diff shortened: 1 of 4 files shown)


therealprof:
Thanks for your PR. I don't think we should apply it though since it will break all applications out there using this functionality without good reason.

Piroro-hs:
I thought that it can be fixed for now as this and other stm32 HALs does not reached to v1.0.0.

therealprof:
> Thanks for fast reply!
> I thought that it can be fixed for now as this and other stm32 HALs does not reached to v1.0.0.

Well, yes and no. 😅 1.0 is just a label and the question is whether those crates will ever reach "1.0" since every change to the svd files or `svd2rust` have to be considered a breaking change and would have to result in an inflationary major version bumping... I'm not opposed to necessary breaking changes (to fix bugs or usability shortcomings) but this particular change seems to lack a good motive.

burrbull:
Rebase, please.

PR #251, opened 2021-01-15
change Serial to use generics instead of macros

xoviat:
This PR modifies the serial module to use generics instead of macros, so that downstream libraries can take advantage of generics. It also includes compatibility functions so that existing code continues to work without modifications.

Feedback is appreciated. The implementation here is ugly, IMO, but the advantage to users of the library is clear.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -1398,360 +1398,422 @@ pub struct Tx<USART> {
     _usart: PhantomData<USART>,
 }
 
-macro_rules! halUsartImpl {
-    ($(
-        $USARTX:ident: ($usartX:ident, $apbXenr:ident, $rcc_bit:expr, $usartXen:ident, $pclkX:ident),
-    )+) => {
-        $(
-            impl<PINS> Serial<$USARTX, PINS> {
-                pub fn $usartX(
-                    usart: $USARTX,
-                    pins: PINS,
-                    config: config::Config,
-                    clocks: Clocks,
-                ) -> Result<Self, config::InvalidConfig>
-                where
-                    PINS: Pins<$USARTX>,
-                {
-                    use self::config::*;
-                    use crate::bb;
+impl<USART, PINS> Serial<USART, PINS>
+where
+    PINS: Pins<USART>,
+    USART: Instance,
+{
+    /*
+        StopBits::STOP0P5 and StopBits::STOP1P5 aren't supported when using UART
 
-                    unsafe {
-                        // NOTE(unsafe) this reference will only be used for atomic writes with no side effects.
-                        let rcc = &(*RCC::ptr());
+        STOP_A::STOP1 and STOP_A::STOP2 will be used, respectively
+    */
+    pub fn new(
+        usart: USART,
+        pins: PINS,
+        config: config::Config,
+        clocks: Clocks,
+    ) -> Result<Self, config::InvalidConfig> {
+        use self::config::*;
 
-                        // Enable clock.
-                        bb::set(&rcc.$apbXenr, $rcc_bit);
-                    }
+        unsafe {
+            // NOTE(unsafe) this reference will only be used for atomic writes with no side effects.
+            let rcc = &(*RCC::ptr());
 
-                    let pclk_freq = clocks.$pclkX().0;
-                    let baud = config.baudrate.0;
+            // Enable clock.
+            // bb::set(&rcc.$apbXenr, $rcc_bit);
+            USART::enable_clock(rcc);
+        }
 
-                    // The frequency to calculate USARTDIV is this:
-                    //
-                    // (Taken from STM32F411xC/E Reference Manual,
-                    // Section 19.3.4, Equation 1)
-                    //
-                    // 16 bit oversample: OVER8 = 0
-                    // 8 bit oversample:  OVER8 = 1
-                    //
-                    // USARTDIV =          (pclk)
-                    //            ------------------------
-                    //            8 x (2 - OVER8) x (baud)
-                    //
-                    // BUT, the USARTDIV has 4 "fractional" bits, which effectively
-                    // means that we need to "correct" the equation as follows:
-                    //
-                    // USARTDIV =      (pclk) * 16
-                    //            ------------------------
-                    //            8 x (2 - OVER8) x (baud)
-                    //
-                    // When OVER8 is enabled, we can only use the lowest three
-                    // fractional bits, so we'll need to shift those last four bits
-                    // right one bit
+        let pclk_freq = USART::pclk_freq(&clocks);
+        let baud = config.baudrate.0;
 
-                    // Calculate correct baudrate divisor on the fly
-                    let (over8, div) = if (pclk_freq / 16) >= baud {
-                        // We have the ability to oversample to 16 bits, take
-                        // advantage of it.
-                        //
-                        // We also add `baud / 2` to the `pclk_freq` to ensure
-                        // rounding of values to the closest scale, rather than the
-                        // floored behavior of normal integer division.
-                        let div = (pclk_freq + (baud / 2)) / baud;
-                        (false, div)
-                    } else if (pclk_freq / 8) >= baud {
-                        // We are close enough to pclk where we can only
-                        // oversample 8.
-                        //
-                        // See note above regarding `baud` and rounding.
-                        let div = ((pclk_freq * 2) + (baud / 2)) / baud;
+        // The frequency to calculate USARTDIV is this:
+        //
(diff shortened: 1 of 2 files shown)


src/serial.rs
```diff
@@ -1398,388 +1399,449 @@ pub struct Tx<USART> {
...
-                    }
-                    Ok(())
+mod private {
+    pub trait Sealed {}
+}
+
+// Implemented by all USART instances
+pub trait Instance: private::Sealed {
+    #[doc(hidden)]
+    fn ptr() -> *const stm32f4::stm32f405::uart4::RegisterBlock;
```
thalesfragoso, line 1785: This will cause problems, we only have that module when using the f405, maybe try to use `crate::pac` instead.

Cargo.toml
```diff
@@ -63,7 +63,7 @@ usbd-serial = "0.1.0"
 micromath = "1.0.0"
 
 [features]
-device-selected = []
+device-selected = ["stm32f4/stm32f405"]
```
thalesfragoso, line 66: This will cause problems if using anything other than the F405. You should only have one mcu feature enabled per compilation.

src/serial.rs
```diff
@@ -1398,388 +1399,449 @@ pub struct Tx<USART> {
...
+
+macro_rules! halUsart {
+    ($(
+        $USARTX:ident: ($usartX:ident, $apbXenr:ident, $rcc_bit:expr, $usartXen:ident, $pclkX:ident),
+    )+) => {
+        $(
+            impl private::Sealed for $USARTX {}
+            impl Instance for $USARTX {
+                fn ptr() -> *const stm32f4::stm32f405::uart4::RegisterBlock {
+                    unsafe { mem::transmute($USARTX::ptr()) }
```
thalesfragoso, line 1802: Can't we get away with just casting pointers ? I prefer to avoid transmute as much as possible. Something like `$USART::ptr() as *const _`

src/serial.rs
```diff
@@ -1398,388 +1399,449 @@ pub struct Tx<USART> {
...
-                        }
-                    }
+                fn pclk_freq(clocks: &Clocks) -> u32 {
+                    clocks.$pclkX().0
                 }
-            }
 
-            impl<PINS> blocking::serial::Write<u8> for Serial<$USARTX, PINS> {
-                type Error = Error;
+                fn enable_clock(rcc: &crate::stm32::rcc::RegisterBlock) {
```
thalesfragoso, line 1809: This should be unsafe.

src/serial.rs
```diff
@@ -1398,388 +1399,449 @@ pub struct Tx<USART> {
...
-                    tx.bwrite_all(bytes)
+                    unsafe { bb::set(&rcc.$apbXenr, $rcc_bit) };
                 }
 
-                fn bflush(&mut self) -> Result<(), Self::Error> {
-                    let mut tx: Tx<$USARTX> = Tx {
-                        _usart: PhantomData,
-                    };
-                    tx.bflush()
+                fn stopbits(bits: config::StopBits) -> stm32f4::stm32f405::uart4::cr2::STOP_A {
```
thalesfragoso, line 1815: Maybe you can get rid of the transmute here if you use `crate::pac` instead of `stm32f4::stm32f405` ?

xoviat: No, because the types are different.

src/serial.rs
```diff
@@ -1398,388 +1399,479 @@ pub struct Tx<USART> {
...
-                where
-                    PINS: Pins<$USARTX>,
-                {
-                    use self::config::*;
-                    use crate::bb;
+// macro_rules! halUsartImpl {
+//     ($(
+//         $USARTX:ident: ($usartX:ident, $apbXenr:ident, $rcc_bit:expr, $usartXen:ident, $pclkX:ident),
+//     )+) => {
+//         $(
```
therealprof, line 1406: Can you get rid of those?

src/serial.rs
```diff
@@ -1398,388 +1399,479 @@ pub struct Tx<USART> {
...
+                    use config::StopBits;
+
+                    unsafe {
+                        mem::transmute(match bits {
+                            StopBits::STOP0P5 => STOP_A::STOP0P5,
+                            StopBits::STOP1 => STOP_A::STOP1,
+                            StopBits::STOP1P5 => STOP_A::STOP1P5,
+                            StopBits::STOP2 => STOP_A::STOP2,
+                        })
+                    }
```
thalesfragoso, line 1856: Can't you use `uart_base` instead of `usart1` to get rid of the `transmute` ?

xoviat: No. uart_base varies between features (I think).

thalesfragoso: Try changing the use to
```rust
use uart_base::cr2::STOP_A;
```

and removing the `transmute`.

thalesfragoso: Actually, I just checked, the enum on `uart4` only has two variants, so this transmute is invalid.

src/serial.rs
```diff
@@ -1398,388 +1399,474 @@ pub struct Tx<USART> {
...
+                    use crate::stm32::usart1::cr2::STOP_A;
+                    use config::StopBits;
+
+                    unsafe {
+                        mem::transmute(match bits {
+                            StopBits::STOP0P5 => STOP_A::STOP0P5,
+                            StopBits::STOP1 => STOP_A::STOP1,
+                            StopBits::STOP1P5 => STOP_A::STOP1P5,
+                            StopBits::STOP2 => STOP_A::STOP2,
+                        })
```
thalesfragoso, line 1850: `uart_base` is sometimes `uart4`, which only has two variants for this enum, so this `transmute` is unsound.

thalesfragoso (changes requested):

xoviat:
@thalesfragoso I removed the transmute.

src/serial.rs
```diff
@@ -1801,27 +1885,54 @@ macro_rules! halUart {
...
+                    use config::StopBits;
+
+                    self
+                        .cr2
+                        .write(|w| w.stop().variant(
+                            match bits {
+                                StopBits::STOP0P5 => STOP_A::STOP1,
+                                StopBits::STOP1 => STOP_A::STOP1,
+                                StopBits::STOP1P5 => STOP_A::STOP2,
+                                StopBits::STOP2 => STOP_A::STOP2,
```
thalesfragoso, line 1915: Could you add docs saying that `StopBits::STOP0P5` and `StopBits::STOP1P5` aren't supported when using UART and which `StopBits` will be used instead ?

thalesfragoso (changes requested):

src/serial.rs
thalesfragoso: Also, could you squash your commits ?

xoviat:
@thalesfragoso Done.

src/serial.rs
```diff
@@ -1398,388 +1398,477 @@ pub struct Tx<USART> {
...
-                        StopBits::STOP1 => STOP_A::STOP1,
-                        StopBits::STOP1P5 => STOP_A::STOP1P5,
-                        StopBits::STOP2 => STOP_A::STOP2,
-                    })
-                });
-                self
+            impl private::Sealed for $USARTX {}
+            impl Instance for $USARTX {
+                fn ptr() -> *const uart_base::RegisterBlock {
+                    $USARTX::ptr() as *const _
```
thalesfragoso, line 1827: I think you can remove this `as *const _`

src/serial.rs
```diff
@@ -1801,27 +1890,54 @@ macro_rules! halUart {
...
-                        StopBits::STOP1 => STOP_A::STOP1,
-                        StopBits::STOP1P5 => STOP_A::STOP2,
-                        StopBits::STOP2 => STOP_A::STOP2,
-                    })
-                });
-                self
+            impl private::Sealed for $USARTX {}
+            impl Instance for $USARTX {
+                fn ptr() -> *const uart_base::RegisterBlock {
+                    $USARTX::ptr() as *const _
```
thalesfragoso, line 1896: Same here

thalesfragoso (commented):
Sorry for the nitpick, but I believe it can make the code more robust with future changes

src/serial.rs
thalesfragoso: Hmm, not sure if this one is possible, due to the weird uarts on the F413

src/serial.rs
thalesfragoso: Nevermind, I got confused

thalesfragoso (approved):
Thank you again and sorry for the delay.

bors r+

PR #254, opened 2021-01-24
Make freeze_unchecked method private

thalesfragoso:
Also add a missing entry for it to the changelog.

This isn't really breaking because this isn't released yet.


therealprof (approved):
LGTM, thanks.

bors r+

PR #255, opened 2021-01-27
Added the self reference to the is_enabled function

justacec:
I believe that the self reference for the DMA Stream method `is_enabled` was inadvertently left out when this was originally authored.  I just added it in.  Not much to say here.


justacec:
Well, looks like I was very wrong about this.  Please disregard this PR.

PR #258, opened 2021-01-31
Add support for DMA with the ADC wrapper.

matoushybl:
Implements the PeriAddress trait for Adc, making it possible for it to be used with the Transfer DMA implementation.

Fixes a rounding problem in the temperature equation. 
Fixes temperature sensor channel assignment for F405. The problem seems to affect all of the F40x and the F41x as the reference manual claims that the channel is 16 instead of 18.


matoushybl:
Hi!
I've added an example of using ADC with RTIC and DMA and rebased the PR.
Would it please be possible to review and merge this?

Thanks!
CC: @therealprof @thalesfragoso

therealprof (approved):
LGTM, thanks!

matoushybl:
Thanks for the review! I've just created a patch that removes the need for the unsafe code, I'll open a new PR with it.

PR #261, opened 2021-02-01
RCC for F412, F413, F423, F446: Add missing configuration of PLLI2SCFGR.PLLI2SN

samcrow:
While testing I2S on an STM32F412, I found that the clock setup code was not correctly configuring the I2S PLL. The clock setup code found a good configuration, but did not write it correctly to the PLL configuration register. The PLLI2SCFGR.PLLI2SN field was left at its default value.

This pull request changes one function to write PLLI2SCFGR.PLLI2SN when applying a PLL configuration.

The updated code is working well for I2S on my STM32F412. I don't have an F413, F423, or F446 to test with, but this seems like a simple fix that is unlikely to break other code.


therealprof (approved):
Good catch. Thanks a lot!

bors r+

PR #262, opened 2021-02-02
CAN: Add more pins, support CAN for more models, add CAN3 where available

samcrow:
I looked through all the STM32F4 datasheets to find additional pins that can be used for CAN, and added implementations of the `can::Pins` trait for those pins. As part of the process, I enabled the `can` module for several STM32F4 models that support CAN beyond the F405, F407, and F446. I also added CAN3 and the associated pins for F413 and F423 models.

This also includes some small documentation improvements that I thought might be helpful.

I'm not sure if this needs a changelog entry because it's in the same version where CAN support was first added. I added one anyway to make the automated checks happy.

### DIFF
--- a/src/can.rs
+++ b/src/can.rs
@@ -140,8 +140,8 @@ macro_rules! bus {
                         // Enable peripheral clock
                         crate::bb::set(&rcc.apb1enr, $peren);
                         // Reset peripheral
-                        crate::bb::set(&rcc.apb2rstr, $peren);
-                        crate::bb::clear(&rcc.apb2rstr, $peren);
+                        crate::bb::set(&rcc.apb1rstr, $peren);
+                        crate::bb::clear(&rcc.apb1rstr, $peren);
                     };
                 }
             }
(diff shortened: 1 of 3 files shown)


src/can.rs
```diff
@@ -1,89 +1,141 @@
...
-            impl sealed::Sealed for crate::pac::$PER {}
-            impl Enable for crate::pac::$PER {
+            impl crate::can::sealed::Sealed for crate::pac::$PER {}
+            impl crate::can::Enable for crate::pac::$PER {
                 #[inline(always)]
                 fn enable() {
                     unsafe {
-                        let rcc = &(*RCC::ptr());
-                        bb::set(&rcc.apb1enr, $peren)
+                        let rcc = &(*crate::pac::RCC::ptr());
```
matoushybl, line 137: Doesn't the peripheral need to be reset too?

matoushybl:
LGTM, thanks for the PR!

src/can.rs
samcrow: I've seen some drivers that reset their peripherals and some that don't. This is the same as how the code worked before I changed it.

Anyway, I do think that resetting the peripheral is a good idea to avoid misconfiguration problems. I'll add a commit for that.

thalesfragoso (approved):
Great work, thanks!

bors r+

PR #263, opened 2021-02-04
RCC: Add option to bypass HSE oscillator

samcrow:
This change adds an option to set the HSEBYP bit of RCC_CR, which disables the built-in oscillator for the HSE clock. Instead, the microcontroller can use a clock signal from an external oscillator or other source.

I have tested this mode on a 32F412GDISCOVERY board.


matoushybl (dismissed):
LGTM, thanks for the PR!

thalesfragoso:
Thank you, could you resolve the conflicts ?

samcrow:
OK, it looks like I've resolved the conflict in the changelog.

thalesfragoso (approved):
Thanks again!

bors r+

PR #264, opened 2021-02-06
RCC for F411: Add missing configuration of PLLI2SCFGR.PLLI2SM

samcrow:
I was testing I2S and noticed that the actual sample rate was wrong (on an STM32F411 this time). I noticed that while the STM32F411 has a separate M divider before the I2S PLL, the clock setup code was not configuring it in the PLLI2SCFGR.

This pull request proposes to change two of the #[cfg] attributes so that the code sets PLLI2SCFGR.PLLI2SM on the STM32F411, like it already does on the F412/413/423/446.

The changed code works and generates the correct sample rate on my STM32F411.


thalesfragoso (approved):
Thanks again!

bors r+

PR #266, opened 2021-02-15
const generic pins

burrbull:
requires Rust 1.51
Depends on #323


therealprof:
CC #272

therealprof:
Sweet! I need to try it out in personally but I do like that very much.

burrbull:
I think we need better names for generic types. Now we have:
```rust
PX<MODE, P, N>,
PXn<MODE, P>, // downgrade
Pin<MODE> // downgrade2
```

therealprof (approved):
Excellent work! Thanks.

bors r+

PR #269, opened 2021-02-27
Clear SPI errors on send to prevent races

riskable:
I'm already aware of #223 but I tried it and...  It doesn't work.  I sort of went back to the drawing board on it and created my own PR that I've tested so I know it works 👍

If errors are encountered during an SPI write it'll automatically clear any bits that need clearing *then* return the error.  So you get the bit cleared and the error.  What more could you ask for? :smile:


therealprof (approved):
LGTM, thanks!

bors r+

PR #271, opened 2021-03-10
PWM Input abstraction

theunkn0wn1:
It also presently depends on side effects﻿, due to my limited experience writing at the HAL layer.
There is also one unsafe block when configuring the prescalers, as it seems (or my IDE is failing me) that the prescaler's don't have safe abstractions for configuring them.

Thnkgs that need be done:
 - [x] get it working for one MCU
 - [x] use macros to generate the definition
 - [x] get it working with all compatible timers across all compatible f4 devices
 - [x] configure the timer's frequency internally instead of depending on side effects
Feedback and suggestions are welcome.
- [x] rework as per #337

### DIFF
--- a/examples/pwm-input.rs
+++ b/examples/pwm-input.rs
@@ -1,0 +1,43 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+// Halt on panic
+use panic_halt as _;
+
+use cortex_m;
+use cortex_m_rt::entry;
+use stm32f4xx_hal::{prelude::*, stm32, timer::Timer};
+
+#[entry]
+fn main() -> ! {
+    if let Some(dp) = stm32::Peripherals::take() {
+        // Set up the system clock.
+        let rcc = dp.RCC.constrain();
+        let clocks = rcc.cfgr.freeze();
+
+        let gpioa = dp.GPIOA.split();
+        let gpioc = dp.GPIOC.split();
+
+        let channels = (gpioa.pa8.into_alternate(), gpioa.pa9.into_alternate());
+        // configure tim1 as a PWM output of known frequency.
+        let pwm = Timer::new(dp.TIM1, &clocks).pwm(channels, 501u32.hz());
+        let (mut ch1, _ch2) = pwm;
+        let max_duty = ch1.get_max_duty();
+        ch1.set_duty(max_duty / 2);
+        ch1.enable();
+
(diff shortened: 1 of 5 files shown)


burrbull:
What status of this?

theunkn0wn1:
The status of this is as follows:
- it works for the F446
- Looks like it fails linting and seems to break on some of other supported F4 chips based on lint errors

This PR slipped off my radar. I will put some more work into it in the immediate future to see if i can't fix the lint errors.

theunkn0wn1:
This should be feature complete.
The only chip this PR doesn't presently support is the F410, which seems to be an oddball with its TIM5 register not having expected fields. (`tim5's `CCR1` register seems to be mising `.ccr`)

examples/pwm-input.rs
```diff
@@ -0,0 +1,48 @@
...
+        ch1.enable();
+        // configure tim8 as a timer, using the best-guess frequency of the input signal.
+        // FIXME: using this for the side effect of configuring the timer, do this ourselves.
+        let tim8 = timer::Timer::tim8(dp.TIM8, 500.hz(), clocks).release();
+
+        // Configure tim8's channel 1 pin into AF3
+        let pwm_reader_ch1 = gpioc.pc6.into_alternate_af3();
+        // instantiate the PWM reader from tim8 and tim8 channel 1.
+
+        let pwm_reader = pwm_input::PwmInput::tim8(tim8, pwm_reader_ch1);
```
burrbull, line 40: I plan to change timers API to be similar to `stm32f1xx-hal`. See #337.
With this change you could rewrite it to:
```rust
let pwm_reader_ch1 = gpioc.pc6.into_alternate();
let pwm_reader = timer::Timer:new(dp.TIM8, clocks).pwm_input(pwm_reader_ch1);
```

theunkn0wn1: Thanks. Ended up rewriting this PR due to #337 but the result is much cleaner this time around.

burrbull:
Looks good to me. Please, squash commits and rebase.

burrbull:
Have you see
https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/pwm_input.rs

theunkn0wn1:
> Have you see
> https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/pwm_input.rs

This is the first time seeing this, honestly I developed this feature without looking for existing examples in other HALs.

burrbull (approved):

PR #272, opened 2021-03-10
Add second downgrade function that removes both pin and port from the type

dimpolo:
I did three things here.

* added a `downgrade2` function to all pins. It returns a `Pin` that is no longer generic over the pin number or its port.
* added a `PinInfo` trait with `pin_id()` and `port_id()` functions
* moved the `ExtiPin` trait implementation out of the macro and into the trait definition since otherwise I would have needed a third `exti` macro

Names are just temporaries.

Why I believe this is necessary:
The current downgrade implementation feels useless to me. If you need to combine pins from different ports you're left with either trait objects or monomorphized generic implementations which both have their downsides.

Are my changes ok?
If so what should the function be called and do we want to remove the current downgrade function?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,11 +52,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added support for CAN on additional models: STM32F412, STM32F413, STM32F415,
   STM32F417, STM32F423, STM32F427, STM32F429, STM32F437, STM32F439, STM32F469,
   and STM32F479 [#262]
+- Added `gpio::gpiox::Pxi::downgrade2` method [#272]
 
 [#231]: https://github.com/stm32-rs/stm32f4xx-hal/pull/231
 [#262]: https://github.com/stm32-rs/stm32f4xx-hal/pull/262
 [#263]: https://github.com/stm32-rs/stm32f4xx-hal/pull/263
 [#278]: https://github.com/stm32-rs/stm32f4xx-hal/issues/278
+[#272]: https://github.com/stm32-rs/stm32f4xx-hal/issues/272
 
 ### Fixed
 
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -2,6 +2,8 @@
 
 use core::marker::PhantomData;
 
+use embedded_hal::digital::v2::{toggleable, InputPin, OutputPin, StatefulOutputPin};
+
 use crate::pac::EXTI;
 use crate::syscfg::SysCfg;
 
@@ -756,6 +758,14 @@ macro_rules! gpio {
                             _mode: self._mode,
                         }
                     }
+
+                    /// Erases the pin number and the port from the type
+                    ///
+                    /// This is useful when you want to collect the pins into an array where you
+                    /// need all the elements to have the same type
+                    pub fn downgrade2(self) -> super::Pin<MODE>{
+                        super::Pin::new($rcc_bit, $i)
+                    }
                 }
 
                 impl<MODE> OutputPin for $PXi<Output<MODE>> {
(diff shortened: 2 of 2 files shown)


dimpolo:
Any Feedback?
Anything I can do to make this easier to review?

therealprof:
Sorry, life is very taxing at the moment so I don't have time to do a proper review and experimentation.

In general I'm good with the intention and we should absolutely have a better downgraded pin implementation. The Rust capabilities are constantly improving and I have a feeling const generics in 1.51 might unlock a totally new improved approach which we might want to take a look at because one thing I'd like to avoid is that we'll have to revisit the whole topic way too soon.

It is a massive change (mostly thanks to the ugly macro salad) and it has conflicts. It would be very helpful if it could be broken down into smaller pieces to allow for a quick review I can sneak in my super busy schedule.

dimpolo:
All good. I don't mean to rush. Glad to hear that the general direction is appreciated.

I'll break up the changes into smaller commits.
You are right about const generics and I'll try to keep this orthogonal to potential changes related to them.

burrbull:
In `stm32f1xx-hal` such enum is used for this:
https://docs.rs/stm32f1xx-hal/0.7.0/stm32f1xx_hal/gpio/enum.Pxx.html
What way is better?

dimpolo:
This design is based on `nrf52`'s implementation: 
[docs](https://docs.rs/nrf52840-hal/0.12.1/nrf52840_hal/gpio/struct.Pin.html) [code](https://github.com/nrf-rs/nrf-hal/blob/04e3fcb03c171c1a68f29e989a0021780992bbc0/nrf-hal-common/src/gpio.rs#L51-L56)
I see two minor problems with the enum way:
* the enum is double the size (`u8` + discriminant)
* when accessing the register `match` might be less efficient than  bitmask + bitshift

burrbull:
>     * the enum is double the size (`u8` + discriminant)

Both variants have same size
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=43b072a223d96de4c47184fc6ed5c222

dimpolo:
A port has at most 16 pins, so we can get away with storing port and pin in a single u8.

https://github.com/stm32-rs/stm32f4xx-hal/blob/8515dec8e1df59eff7a899c839a2eb73d3704abb/src/gpio.rs#L1165-L1177

burrbull:
> A port has at most 16 pins, so we can get away with storing port and pin in a single u8.
> 
> https://github.com/stm32-rs/stm32f4xx-hal/blob/8515dec8e1df59eff7a899c839a2eb73d3704abb/src/gpio.rs#L1165-L1177

Interesting idea. But in this case you pay with code size and speed when switch/read pin state.

burrbull:
I'm not mind your way, but I want to see real proofs that packing is better.

dimpolo:
I rebased, changed some things and left the EXTI implementation for later to not mess with the macros for now.
Code should hopefully be nicer to review.

@burrbull how could I aproach a comparison?  See what method toggles a pin faster? Dissasembly?

burrbull:
> how could I aproach a comparison? See what method toggles a pin faster? Dissasembly?

I think it will be enough to compare firmware size (see `cargo binutils`).
I'd like to see comparison of `Pin` structure with and without packing pin_id and port_id.
For example. Collect 10 different pin in array and pass them to function marked with `#[inline(never)]`.
In function change pin states randomly several times.

dimpolo:
Something like this?
https://github.com/dimpolo/stm32f4xx-hal/blob/comparisons/examples/downgrade_comparison.rs

burrbull:
@dimpolo Rebase, please. And is this still draft?

burrbull (approved):

PR #277, opened 2021-03-19
Update SDIO_D0 pin to PB7 for stm32f411

antonok-edm:
As per page 46 of https://www.st.com/resource/en/datasheet/stm32f411ce.pdf, the `SDIO_D0` pin option is actually on PB7, not `PB6`.

The `PB6` assignment appears to be correct for the other three microcontrollers.

### DIFF
--- a/src/sdio.rs
+++ b/src/sdio.rs
@@ -94,16 +94,21 @@ pins! {
     D3: [PC11<Alternate<AF12>>]
 }
 
-#[cfg(any(
-    feature = "stm32f411",
-    feature = "stm32f412",
-    feature = "stm32f413",
-    feature = "stm32f423",
-))]
+#[cfg(any(feature = "stm32f412", feature = "stm32f413", feature = "stm32f423"))]
 pins! {
     CLK: [PB15<Alternate<AF12>>]
     CMD: [PA6<Alternate<AF12>>]
     D0: [PB4<Alternate<AF12>>, PB6<Alternate<AF12>>]
+    D1: [PA8<Alternate<AF12>>]
+    D2: [PA9<Alternate<AF12>>]
+    D3: [PB5<Alternate<AF12>>]
+}
+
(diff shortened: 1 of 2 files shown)


therealprof:
Interesting. I can confirm your finding looking at https://github.com/STMicroelectronics/STM32_open_pin_data/blob/master/mcu/STM32F411V(C-E)Hx.xml

In order to make that compile you will have to duplicate the while `pins!` block though, you can't single out the `D0` signal.

therealprof:
Uhm, that doesn't look right. 😅

I think you need to copy the full block, remove the `stm32f411` case from the cfg gate and remove `PB7` from the `D0` assignment, and then cfg gate the new block with just with `stm32f411` and just keep`PB7`.

Also please move the Changelog entry under the Changes category.

src/sdio.rs
```diff
@@ -103,12 +103,32 @@ pins! {
...
+}
+
+#[cfg(any(feature = "stm32f412", feature = "stm32f413", feature = "stm32f423"))]
+pins! {
+    CLK: []
+    CMD: []
+    D0: [PB6<Alternate<AF12>>]
+    D1: []
+    D2: []
+    D3: []
```
antonok-edm, line 129: I just added the rest of the macro and added a changelog entry as well.

antonok-edm:
ok, I think I got it this time? :sweat_smile:

therealprof:
Almost, I see you also added `PB4` (which seems to be correct)? Want to mention that in the Changelog as well? 😉

antonok-edm:
`PB4` is the same as on the other 3 microcontrollers and already existed for `stm32f411` on the previous implementation, so I figured that isn't really part of the change here.

therealprof (approved):
Ah, my bad. I misread that as PB6 or PB7. LGTM then. Thanks a lot!

bors r+

PR #283, opened 2021-03-25
Fix CRC32 peripheral

### LINKED ISSUE #281: CRC32 always emits zero when fed bytes
MCU: stm32f446RETx

It appears the CRC peripheral always emits a computed value of zero, regardless of the bytes feed to it.

reproducible sample
```rs
#![deny(unsafe_code)]
#![no_main]
#![no_std]

// panic via probe
use panic_probe as _;

use cortex_m;
use cortex_m_rt::entry;
use stm32f4xx_hal::{prelude::*, crc32, stm32};

use rtt_target::{rtt_init_print, rprintln};

static BUFFER: [u8; 4] = [0x32u8, 0x42u8, 0x12u8, 0x44u8];

#[entry]
fn main() -> ! {
    rtt_init_print!();
    rprintln!("hello, world!");
    if let Some(dp) = stm32::Peripherals::take() {
        // Set up the system clock.
        let rcc = dp.RCC.constrain();

        // Initialize the CRC32 peripheral
        let mut crc = crc32::Crc32::new(dp.CRC);
        // declare a buffer
        let result = crc.update_bytes(&BUFFER);
        rprintln!("crc32 of {:?} = {}", BUFFER, result);
    }

    loop {
        cortex_m::asm::nop();
    }
}
```
Expected result:
```
hello, world!
crc32 of [50, 66, 18, 68] = 2252567011
```
Actual Result:
```
hello, world!
crc32 of [50, 66, 18, 68] = 0
```

theunkn0wn1:
The current implementation of the CRC32 perhipheral does not enable the CRC32 clock.
On the STM32F446﻿, this causes **all** calculations against this peripheral to emit an incorrect null instead of the checksum.

This PR enables the CRC clock when the HAL's peripheral is acquired, and disables it when released.

Fixes #281

### DIFF
--- a/src/crc32.rs
+++ b/src/crc32.rs
@@ -21,16 +21,16 @@ pub struct Crc32 {
 impl Crc32 {
     /// Create a new Crc32 HAL peripheral
     pub fn new(crc: CRC) -> Self {
-        let mut new = Self { periph: crc };
-        new.init();
-
         unsafe {
             // NOTE(unsafe) this reference will only be used for atomic writes with no side effects.
-            let rcc_raw =  &(*RCC::ptr());
+            let rcc_raw = &(*RCC::ptr());
             // CRCEN = true; enable CRC clock.
-            bb::clear(&rcc_raw.ahb1enr, 12);
+            bb::set(&rcc_raw.ahb1enr, 12);
             dsb();
         }
+
+        let mut new = Self { periph: crc };
+        new.init();
 
         new
     }
@@ -123,9 +123,9 @@ impl Crc32 {
     pub fn free(self) -> CRC {
         unsafe {
             // NOTE(unsafe) this reference will only be used for atomic writes with no side effects.
-            let rcc_raw =  &(*RCC::ptr());
+            let rcc_raw = &(*RCC::ptr());
             // Disable CRC clock
-            bb::set(&rcc_raw.ahb1enr, 12);
+            bb::clear(&rcc_raw.ahb1enr, 12);
             dsb();
         }
 
--- a/src/crc32.rs
+++ b/src/crc32.rs
@@ -7,9 +7,10 @@
 //! It operates word-at-a-time, and takes 4 AHB/HCLK cycles per word
 //! to calculate. This operation stalls the AHB bus for that time.
 
-use crate::stm32::CRC;
+use crate::stm32::{CRC, RCC};
 use core::mem::MaybeUninit;
 use core::ptr::copy_nonoverlapping;
+use cortex_m::asm::{nop, delay};
 
 /// A handle to a HAL CRC32 peripheral
 pub struct Crc32 {
@@ -21,6 +22,14 @@ impl Crc32 {
     pub fn new(crc: CRC) -> Self {
         let mut new = Self { periph: crc };
         new.init();
+
+        unsafe {
+            // NOTE(unsafe) this reference will only be used for atomic writes with no side effects.
+            let rcc_raw = unsafe { &(*RCC::ptr()) };
+            // Enable CRC clock
+            rcc_raw.ahb1enr.modify(|_, w| {w.crcen().enabled()});
+        }
+
         new
     }
 
@@ -39,7 +48,6 @@ impl Crc32 {
         for word in data {
             self.periph.dr.write(|w| unsafe { w.bits(*word) });
         }
-
         // Retrieve the resulting CRC
         self.periph.dr.read().bits()
     }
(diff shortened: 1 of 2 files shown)


src/crc32.rs
```diff
@@ -21,6 +21,14 @@ impl Crc32 {
     pub fn new(crc: CRC) -> Self {
         let mut new = Self { periph: crc };
         new.init();
+
+        unsafe {
+            // NOTE(unsafe) this reference will only be used for atomic writes with no side effects.
+            let rcc_raw = unsafe { &(*RCC::ptr()) };
+            // Enable CRC clock
+            rcc_raw.ahb1enr.modify(|_, w| w.crcen().enabled());
```
therealprof, line 29: Please use bitbanding to change the flag instead, cf. `gpio.rs`.

Also after enabling the clocks, you might want to throw in a `cortex_m::asm::dsb();` for good measure to work around an erratum, also cf. `gpio.rs`.

theunkn0wn1: Done.

src/crc32.rs
```diff
@@ -111,6 +121,14 @@ impl Crc32 {
 
     /// Consume the HAL peripheral, returning the PAC peripheral
     pub fn free(self) -> CRC {
+        unsafe {
+            // NOTE(unsafe) this reference will only be used for atomic writes with no side effects.
+            let rcc_raw = unsafe { &(*RCC::ptr()) };
```
thalesfragoso, line 126: This is already inside a unsafe block, so the second unsafe is unnecessary and emits a warning.

src/crc32.rs
```diff
@@ -111,6 +121,14 @@ impl Crc32 {
...
+            // Disable CRC clock
+            bb::write(&rcc_raw.ahb1enr, 12, false);
```
thalesfragoso, line 128: Might be better to use `bb::clear` ?

src/crc32.rs
```diff
@@ -21,6 +23,15 @@ impl Crc32 {
     pub fn new(crc: CRC) -> Self {
         let mut new = Self { periph: crc };
         new.init();
+
+        unsafe {
+            // NOTE(unsafe) this reference will only be used for atomic writes with no side effects.
+            let rcc_raw = unsafe { &(*RCC::ptr()) };
```
thalesfragoso, line 29: This is already inside a unsafe block, so the second unsafe is unnecessary and emits a warning.

src/crc32.rs
```diff
@@ -21,6 +23,15 @@ impl Crc32 {
...
+            // CRCEN = true; enable CRC clock.
+            bb::write(&rcc_raw.ahb1enr, 12, true);
```
thalesfragoso, line 31: Might be better to use `bb::set` ?

src/crc32.rs
```diff
@@ -21,6 +23,15 @@ impl Crc32 {
     pub fn new(crc: CRC) -> Self {
         let mut new = Self { periph: crc };
         new.init();
```
thalesfragoso, line 25: You probably want to enable the clocks before this.

thalesfragoso (changes requested):
Thanks for spotting the problem and providing a solution! I left a few suggestions.

src/crc32.rs
theunkn0wn1: Probably right. Adjusted.

src/crc32.rs
theunkn0wn1: Oops! will correct.

src/crc32.rs
theunkn0wn1: Moved to be the first thing `Crc32::new` does. I don't believe it makes a considerable difference as calculations arn't done until the first write to the data register; which occurs outside this constructor.

src/crc32.rs
theunkn0wn1: Corrected.

src/crc32.rs
thalesfragoso: Thanks, the difference was that the `init` was resetting an internal state, which probably wouldn't do anything with the clocks disabled.

theunkn0wn1: The implementation of `init` sets the `RESET` bit of `CRC_CR`, which is automatically cleared by the hardware.
Even if the reset doesn't occur until after the clock is enabled, I suspect the device would still be reset.

Either way, its a fair point and I have no problem enabling the clock earlier.

thalesfragoso (approved):
Thanks again!

bors r+

PR #284, opened 2021-03-29
qei: implement generics

xoviat:
- allows use of new function for any timer
- deprecates timX functions


therealprof (approved):
LGTM, thanks.

bors r+

PR #294, opened 2021-04-10
Updates and cleanups to match the changes in the PAC

jkristell:
With this crates update of the PAC dependency came my PAC patches to the sdio peripheral. This PR adjust the sdio driver to take advantage of those changes.

This should get rid of the warnings about unnecessary unsafes in the sdio driver.


thalesfragoso (approved):
Thanks!

bors r+

PR #300, opened 2021-04-16
Fix duplicate "Added" sections in changelog

samcrow:
With three recently merged pull requests, we ended up with two "Added" sections in different parts of the unreleased section of the changelog. I propose to combine them into one section, above the "Changed" section.


therealprof (approved):
Hahaha, thanks. And I was already kind of curious why we didn't get the obligatory merge conflict...

bors r+

PR #301, opened 2021-04-18
Remove unsafe code from the ADC DMA example.

matoushybl:


thalesfragoso (changes requested):
Thanks for the PR, but one thing, the DMA config enables double buffering but you're only providing one buffer to `Transfer::init`, this code will panic. If you want to remove `next_transfer_with` you will need 3 buffers for double buffering.

matoushybl:
I disabled double buffering in the DMA config. Is it ok now?
Thanks for the comment, I've missed it.

thalesfragoso:
Yes, it should be okay now, have you tried it on hardware ?

matoushybl:
I've tried running it out of the HAL project in another with defmt and it worked, but I failed to make it run as an example in the project.
My gdb skills are terrible so I believe that that may have been a problem, or something with the project config.

thalesfragoso:
Weird, but I don't see any problem with the code, so...

thalesfragoso (approved):
Thanks again!

bors r+

PR #303, opened 2021-04-20
Better documentation of rng and prelude

jounathaen:
This adds doc comments to the rng and prelude module as well as to the re-export of the svd2rust API.

The rng documentation might repeat some information, but as the generated documentation splits this to multiple pages, I find it is better that way. I also added small example code to the documentation. The examples in the repository are nice, but I personally would favor examples inside the docs.

It might be good, if somebody reads over the prelude part to check that I'm not completely off-track there.


jounathaen:
Ok, apparently a changelog entry is missing. Is this necessary in this case?

therealprof:
> Ok, apparently a changelog entry is missing. Is this necessary in this case?

Why not? Improved documentation is noteworthy.

jounathaen:
Ok, from my side this is ready for review.
I hope this is acceptable and helps somebody. Thank you all for your work on this project!
:slightly_smiling_face:

burrbull:
Rebase, please.

jounathaen:
Rebased!

burrbull (approved):

PR #306, opened 2021-05-06
derive Copy/Clone and Eq/PartialEq for Error types

dimpolo:


burrbull:
`PartialEq` and `Eq` are also useful.

dimpolo:
Added `PartialEq` and `Eq`

burrbull (approved):

PR #309, opened 2021-05-22
Add delay impls for TIM2/TIM5

dbrgn:
Sometimes SYST cannot be used for delays, for example in the case of
RTIC where SYST is already used by the scheduler. For such cases, this
commit adds `DelayMs` and `DelayUs` trait impls (both for u32 and u16)
based on the 32-bit general purpose TIM2/TIM5 timers.

The implementation uses one-pulse mode, so that the timer doesn't need
to be explicitly stopped: The CEN flag is automatically cleared once the
auto-reload value in the ARR is reached.

For backwards compatibility, the old `Delay` struct was _not_ renamed to
`SystickDelay`, but that might be something to consider.

Tested with a MiniF4 (aka Black Pill) devboard.

### DIFF
--- a/src/delay/mod.rs
+++ b/src/delay/mod.rs
@@ -3,3 +3,46 @@
 mod syst;
 
 pub use syst::SystickDelay as Delay;
+
+mod timer;
+
+#[cfg(any(
+    feature = "stm32f401",
+    feature = "stm32f405",
+    feature = "stm32f407",
+    feature = "stm32f411",
+    feature = "stm32f412",
+    feature = "stm32f413",
+    feature = "stm32f415",
+    feature = "stm32f417",
+    feature = "stm32f423",
+    feature = "stm32f427",
+    feature = "stm32f429",
+    feature = "stm32f437",
+    feature = "stm32f439",
+    feature = "stm32f446",
+    feature = "stm32f469",
+    feature = "stm32f479"
+))]
+pub use timer::Tim2Delay;
+
+#[cfg(any(
+    feature = "stm32f401",
+    feature = "stm32f405",
+    feature = "stm32f407",
+    feature = "stm32f410",
+    feature = "stm32f411",
+    feature = "stm32f412",
+    feature = "stm32f413",
+    feature = "stm32f415",
+    feature = "stm32f417",
+    feature = "stm32f423",
+    feature = "stm32f427",
+    feature = "stm32f429",
+    feature = "stm32f437",
+    feature = "stm32f439",
+    feature = "stm32f446",
+    feature = "stm32f469",
+    feature = "stm32f479"
+))]
+pub use timer::Tim5Delay;
--- a/src/delay/mod.rs
+++ b/src/delay/mod.rs
@@ -1,0 +1,5 @@
+//! Delays
+
+mod syst;
+
+pub use syst::SystickDelay as Delay;
--- a/src/delay/timer.rs
+++ b/src/delay/timer.rs
@@ -1,0 +1,193 @@
+//! Delay implementation based on general-purpose 32 bit timers.
+//!
+//! TIM2 and TIM5 are a general purpose 32-bit auto-reload up/downcounter with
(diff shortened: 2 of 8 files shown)


src/delay/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    feature = "stm32f417",
+    feature = "stm32f423",
+    feature = "stm32f427",
+    feature = "stm32f429",
+    feature = "stm32f437",
+    feature = "stm32f439",
+    feature = "stm32f446",
+    feature = "stm32f469",
+    feature = "stm32f479"
+))]
```
dbrgn, line 26: I re-used the feature gates from the `timer` module (for the TIM2 / TIM5 timer impl).

src/delay/timer.rs
```diff
@@ -0,0 +1,193 @@
...
+            }
+
+            fn $waitfn(tim: &mut pac::$TIM, prescaler: u16, auto_reload_register: u32) {
+                // Write Prescaler (PSC)
+                tim.psc.write(|w| w.psc().bits(prescaler));
+
+                // Write Auto-Reload Register (ARR)
+                // Note: Make it impossible to set the ARR value to 0, since this
+                // would cause an infinite loop.
+                tim.arr.write(|w| unsafe { w.bits(max(1, auto_reload_register)) });
```
dbrgn, line 33: Note: This `max` call makes it impossible to delay for 0 ms / us, but I don't see the use case for that. It would be undefined (in terms of actual delay time) anyways.

src/delay/timer.rs
```diff
@@ -0,0 +1,193 @@
...
+                        // Enable timer peripheral in RCC
+                        bb::set(&rcc.$apbenr, $en_bit);
+
+                        // Stall the pipeline to work around erratum 2.1.13 (DM00037591)
+                        cortex_m::asm::dsb();
+
+                        // Reset timer
+                        bb::set(&rcc.$apbrstr, $en_bit);
+                        bb::clear(&rcc.$apbrstr, $en_bit);
+                    }
```
dbrgn, line 65: These lines are re-used a few times within the codebase, not sure if it might make sense to create an internal helper function for them...

src/delay/timer.rs
```diff
@@ -0,0 +1,193 @@
...
+
+                    // Set up prescaler so that a tick takes exactly 0.5 ms.
+                    //
+                    // For example, if the clock is set to 48 MHz, with a prescaler of 24'000
+                    // we'll get ticks that are 0.5 ms long. This means that we can write the
+                    // delay value multipled by two to the auto-reload register (ARR).
+                    //
+                    // Note that we cannot simply use a prescaler value where the tick corresponds
+                    // to 1 ms, because then a clock of 100 MHz would correspond to a prescaler
+                    // value of 100'000, which doesn't fit in the 16-bit PSC register.
```
dbrgn, line 121: I'm not 100% sure if APB1 follows the system clock, but I think there are F4 chips that support 100 MHz or more.

dbrgn: Uh, actually there are models (e.g. STM32F469) that support system clocks up to 180 MHz. In that case, the even with a prescaler for 0.5 ms, the value would be 90'000, which doesn't fit in 16 bit.

Any ideas what to do? Should we use a prescaler for 0.25 ms? That would reduce the usable range to (2^32)/4, which is still 12 days for 32 bit values. I think that should be plenty.

Alternatively we could dynamically calculate the PSC and ARR, at the cost of either the range or the resolution. I'd usually prefer resolution over range for a generic implementation. However, a dynamic implementation that might panic for certain values depending on the system clock configuration sucks.

burrbull:
```rust
pub struct Delay<T = SYST> {
    clocks: Clocks,
    tim: T,
}

impl Delay<SYST> {
    ...
}
```

In example:
```rust
let mut delay = hal::delay::Delay::new(dp.TIM5, clocks);
```

PR #311, opened 2021-05-24
Rcc Enable/Reset + features

burrbull:


thalesfragoso:
Hmm, seems like you're removing bit banding, but still not requiring the user to pass in the RCC, and then using `modify` which might cause races, if we want to go ahead with this we might want to use a critical section or change the traits implementations to use bit banding.

burrbull:
> Hmm, seems like you're removing bit banding, but still not requiring the user to pass in the RCC, and then using `modify` which might cause races, if we want to go ahead with this we might want to use a critical section or change the traits implementations to use bit banding.

Revert bit-banding. See last commit.

src/pwm.rs
```diff
@@ -229,12 +210,12 @@ macro_rules! pwm_all_channels {
 
                 //NOTE(unsafe) atomic write with no side effects
                 fn disable(&mut self) {
-                    unsafe { bb::clear(&(*$TIMX::ptr()).ccer, 0) }
+                    unsafe { (*$TIMX::ptr()).ccer.modify(|_,w| w.cc1e().clear_bit()); }
```
thalesfragoso, line 213: Sorry, I missed that before, I used bit-banding here because a channel can race with another one while modifying this register. This is also the case for the other `disable`s/`enable`s you changed here.

src/rtc.rs
```diff
@@ -416,38 +414,21 @@ fn hours_to_u8(hours: Hours) -> Result<u8, Error> {
...
-    }
+    // Force a reset of the backup domain.
+    rcc.bdcr.modify(|_, w| w.bdrst().set_bit());
+    rcc.bdcr.modify(|_, w| w.bdrst().clear_bit());
+    // Enable the LSE.
+    rcc.bdcr
+        .modify(|_, w| w.lseon().set_bit().lsebyp().bit(bypass));
+    while rcc.bdcr.read().lserdy().bit_is_clear() {}
+    // Set clock source to LSE.
+    rcc.bdcr.modify(|_, w| w.rtcsel().lse());
```
thalesfragoso, line 425: This was also using `bb` to avoid races, that `&RegisterBlock` is acquired with `unsafe`

src/rtc.rs
```diff
@@ -459,8 +440,5 @@ fn unlock(rcc: &RegisterBlock, pwr: &mut PWR) {
 
 fn enable(rcc: &RegisterBlock) {
     // Start the actual RTC.
-    // Set BDCR - Bit 15 (RTCEN)
-    unsafe {
-        bb::set(&rcc.bdcr, 15);
-    }
+    rcc.bdcr.modify(|_, w| w.rtcen().set_bit());
```
thalesfragoso, line 443: This now can also race.

thalesfragoso (changes requested):
Really outstanding work, I just noticed a few small problems, but thank you!

thalesfragoso (approved):
Thanks again!

PR #318, opened 2021-06-18
adds missing dma error flags

peauters:
This PR adds the missing methods to retrieve the error flags for dma streams.

PR #319, opened 2021-06-18
adds dma support for spi

peauters:
This is a first draft of what DMA support for SPI would look like.

I don't have a great opinion on how some of the error flags can/should be used in conjunction with DMA transfers and would love some advice.

Once #318 has been merged, I'll update the example to look at the extra error flags to show multiple transfers.

### DIFF
--- a/examples/spi_dma.rs
+++ b/examples/spi_dma.rs
@@ -1,0 +1,115 @@
+#![no_main]
+#![no_std]
+
+// Halt on panic
+use panic_halt as _;
+
+use core::cell::RefCell;
+use cortex_m::interrupt::Mutex;
+use cortex_m_rt::entry;
+use embedded_hal::spi::{Mode, Phase, Polarity};
+use stm32f4xx_hal::pac::interrupt;
+use stm32f4xx_hal::{
+    dma::{config, traits::StreamISR, MemoryToPeripheral, Stream4, StreamsTuple, Transfer},
+    pac,
+    prelude::*,
+    spi::*,
+};
+
+const ARRAY_SIZE: usize = 100;
+
+type SpiDma = Transfer<
+    Stream4<pac::DMA1>,
+    Tx<pac::SPI2>,
+    MemoryToPeripheral,
+    &'static mut [u8; ARRAY_SIZE],
+    0,
+>;
+
+static G_TRANSFER: Mutex<RefCell<Option<SpiDma>>> = Mutex::new(RefCell::new(None));
+
+#[entry]
+fn main() -> ! {
+    if let Some(dp) = pac::Peripherals::take() {
+        // Set up the system clock.
+        let rcc = dp.RCC.constrain();
+        let clocks = rcc.cfgr.freeze();
+
+        let steams = StreamsTuple::new(dp.DMA1);
+        let stream = steams.4;
+
+        let gpiob = dp.GPIOB.split();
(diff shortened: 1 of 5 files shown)


src/spi.rs
```diff
@@ -1096,6 +1102,77 @@ where
...
+
+pub struct Tx<SPI> {
+    spi: PhantomData<SPI>,
+    dr: u32,
+}
+
+pub struct Rx<SPI> {
+    spi: PhantomData<SPI>,
+    dr: u32,
+}
```
thalesfragoso, line 1117: If you bound this `SPI` to an `Instance` trait (or other name) with a `fn dr() -> u32` method, you could remove this `dr` field and save some ram. You could then implement this `Instance` trait for `SPI1` (and others) with something like:
```
unsafe { &(&*SP1::ptr().dr) as *const _ as u32 }
```

peauters: Sounds good.

peauters:
I've made the change suggested by Thales.

I'm having trouble getting the example to work with two transfers. It only transfers once. I have it working in my project using rtic, so my guess is that there is some concurrency going on that is whooshing above my head. Any got any ideas what I'm missing?

examples/spi_dma.rs
```diff
@@ -0,0 +1,109 @@
...
+                .transfer_complete_interrupt(true),
+        );
+
+        transfer.start(|_tx| {});
+
+        // Hand off transfer to interrupt handler
+        cortex_m::interrupt::free(|cs| *G_TRANSFER.borrow(cs).borrow_mut() = Some(transfer));
+        // Enable interrupt
+        unsafe {
+            cortex_m::peripheral::NVIC::unmask(stm32::Interrupt::DMA1_STREAM5);
```
kalkyl, line 78:
```suggestion
            cortex_m::peripheral::NVIC::unmask(stm32::Interrupt::DMA2_STREAM4);
```

src/spi.rs
```diff
@@ -781,6 +824,69 @@ where
...
+    pub fn rx(self) -> Rx<SPI> {
+        self.new_rx()
+    }
+
+    pub fn txrx(self) -> (Tx<SPI>, Rx<SPI>) {
+        (self.new_tx(), self.new_rx())
+    }
+
+    fn new_tx(&self) -> Tx<SPI> {
+        self.spi.cr2.write(|w| w.txdmaen().enabled());
```
kalkyl, line 856:
```suggestion
        self.spi.cr2.modify(|_, w| w.txdmaen().enabled());
```

src/spi.rs
```diff
@@ -781,6 +824,69 @@ where
...
+        Tx { spi: PhantomData }
+    }
+
+    fn new_rx(self) -> Rx<SPI> {
+        self.spi.cr2.write(|w| w.rxdmaen().enabled());
```
kalkyl, line 861:
```suggestion
        self.spi.cr2.modify(|_, w| w.rxdmaen().enabled());
```
... or else txrx() doesn't work because txdmaen is reset

burrbull:
cc @peauters

peauters:
Apologies for leaving this open for so long. 

I got diverted off onto another project. 

I'll look to get this in to shape in the next week or so.

kalkyl:
Are there any showstoppers left for merging this one?

burrbull:
@kalkyl  Can you test this?

burrbull (approved):

PR #327, opened 2021-07-03
serial pins post-configure

burrbull:
Depends on #334


burrbull:
Closed in favour of #359

PR #338, opened 2021-07-08
use embedded-hal 1.0 with ehal1 feature

burrbull:


no111u3:
Code looks like C with many ‘ifdef’. Might be better to split another version of instances to another files in one module

burrbull:
Closed in favor of #388

PR #339, opened 2021-07-10
Change generic pin names

burrbull:


therealprof (approved):
LGTM in general.

I think it would be nice to have some documentation about what erased and partially erased means but we can do that in a follow-up.

burrbull:
bors r=therealprof

PR #349, opened 2021-07-17
Add bidirectional mode for spi

no111u3:
Add bidirectional mode for spi. As previous pr #210 this pr provide bidi mode for spi but more confidence and feature full.


burrbull:
About error flags https://github.com/stm32-rs/stm32f4xx-hal/pull/223

burrbull:
It would be nice to have example describing how to use different modes.

no111u3:
> About error flags #223

Fixed.

no111u3:
> It would be nice to have example describing how to use different modes.

Will be added later

burrbull (approved):

no111u3:
> It would be nice to have example describing how to use different modes.
Add simple example witch use SPI BiDi mode for transfer data.

burrbull (approved):

PR #357, opened 2021-08-06
Add listen/unlisten for RX/TX-only USART

kalkyl:
This PR adds `listen` and `unlisten` functions for the RX and TX only USART modes.
It also adds functions for checking the relevant interrupt flags and for clearing the idle line interrupt flag.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -543,25 +543,30 @@ where
         }
     }
 
-    /// Start listening for an interrupt event
+    /// Start listening for an rx not empty interrupt event
     ///
     /// Note, you will also have to enable the corresponding interrupt
     /// in the NVIC to start receiving events.
-    pub fn listen(&mut self, event: Event) {
-        match event {
-            Event::Rxne => unsafe { (*USART::ptr()).cr1.modify(|_, w| w.rxneie().set_bit()) },
-            Event::Txe => unsafe { (*USART::ptr()).cr1.modify(|_, w| w.txeie().set_bit()) },
-            Event::Idle => unsafe { (*USART::ptr()).cr1.modify(|_, w| w.idleie().set_bit()) },
-        }
+    pub fn listen(&mut self) {
+        unsafe { (*USART::ptr()).cr1.modify(|_, w| w.rxneie().set_bit()) }
     }
 
-    /// Stop listening for an interrupt event
-    pub fn unlisten(&mut self, event: Event) {
-        match event {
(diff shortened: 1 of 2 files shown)


burrbull:
I'm not sure it is good idea to give access to `listen(Txe)` for `Rx` and vice versa.

kalkyl:
> I'm not sure it is good idea to give access to `listen(Txe)` for `Rx` and vice versa.

Would you prefer having separate functions like: `enable_idle_interrupt()`
`disable_idle_interrupt()` 
`enable_rxne_interrupt()`
`disable_rxne_interrupt()`
`enable_txne_interrupt()`
`disable_txne_interrupt()`

burrbull:
```rust
impl Rx {
    pub fn listen(&mut self) {
        unsafe { (*USART::ptr()).cr1.modify(|_, w| w.rxneie().set_bit()) }
    }
    pub fn unlisten(&mut self) { .. }
    pub fn listen_idle(&mut self) {
         unsafe { (*USART::ptr()).cr1.modify(|_, w| w.idleie().set_bit()) }
    }
    pub fn unlisten_idle(&mut self) { .. }
    pub fn is_idle(&self) { .. }
    pub fn is_rx_not_empty(&self) { .. }
}
impl Tx {
    pub fn listen(&mut self) { .. }
    pub fn unlisten(&mut self) { .. }
    pub fn is_tx_empty(&self) { .. } 
}

**Comment by burrbull** [comment]
Could you rename them to more verbose `is_tx_empty` and `is_rx_not_empty`?

File: src/serial.rs [hunk 1]
```diff
@@ -750,6 +818,14 @@ where
         unsafe { (*USART::ptr()).sr.read().rxne().bit_is_set() }
     }
 
+    // Clear idle line interrupt flag
+    pub fn clear_idle_interrupt(&self) {
+        unsafe {
+            let _ = (*USART::ptr()).sr.read().idle();
+            let _ = (*USART::ptr()).dr.read().bits();
```
burrbull, line 825:
```rust
let _ = (*USART::ptr()).sr.read();
let _ = (*USART::ptr()).dr.read();
```
is enough here.

burrbull:
> > Could you rename `is_txe` and `is_rxne` to more verbose `is_tx_empty` and `is_rx_not_empty`?
> 
> Yes but i thought it would make more sense if they were named like they are in `serial`? Should we rename those too and have a breaking change you think?

We can add new methods to `Serial` and mark old as deprecated.

burrbull (approved):

burrbull:
bors merge

PR #359, opened 2021-08-14
SetAlternate

burrbull:
With this PR passing `into_alternate()` is not necessary (but still can be used) when you work with `Serial`, `Spi`, `I2c` or `I2s`.
Pins will be internally switched in alternated mode and restored on `release`.

### DIFF
--- a/src/fmpi2c.rs
+++ b/src/fmpi2c.rs
@@ -1,7 +1,7 @@
 use core::ops::Deref;
 use embedded_hal::blocking::i2c::{Read, Write, WriteRead};
 
-use crate::gpio::{gpiob, gpioc, gpiod, gpiof, AlternateOD};
+use crate::gpio::{gpiob, gpioc, gpiod, gpiof, Const, SetAlternateOD};
 use crate::i2c::{Error, PinScl, PinSda};
 use crate::pac::{fmpi2c1, FMPI2C1, RCC};
 use crate::rcc::{Enable, Reset};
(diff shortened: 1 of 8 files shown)


src/fmpi2c.rs
```diff
@@ -64,25 +64,33 @@ where
...
+
+impl<SCL, SDA, const SCLA: u8, const SDAA: u8> FMPI2c<FMPI2C1, (SCL, SDA)>
 where
-    SCL: PinScl<FMPI2C1>,
-    SDA: PinSda<FMPI2C1>,
+    SCL: PinScl<FMPI2C1, A = Const<SCLA>> + SetAlternateOD<SCLA>,
+    SDA: PinSda<FMPI2C1, A = Const<SDAA>> + SetAlternateOD<SDAA>,
 {
-    pub fn new<M: Into<FmpMode>>(i2c: FMPI2C1, pins: (SCL, SDA), mode: M) -> Self {
+    pub fn new<M: Into<FmpMode>>(i2c: FMPI2C1, mut pins: (SCL, SDA), mode: M) -> Self {
```
therealprof, line 93: Since you've been mentioning breaking changes in the CHANGELOG, why is this not considered a breaking change?

burrbull: Are you asking about all PR? If so, I've not found it breaks something that had worked earlier.

If you are about  this `mut`. It is internal `mut`. It does not affect on API.

therealprof: Hm, it is a public method so someone could use it, no?

burrbull: This is not type `mut`. This is value `mut`. Inside function.

burrbull:
```rust
fn new(mut pins: (SCL, SDA)) {
    // do something with `pins`
}
```
is just shorthand for:
```rust
fn new(pins: (SCL, SDA)) {
    let mut pins = pins;
    // do something with `pins`
}
```

therealprof: 🤦‍♂️ Somehow I was seeing a reference there. Never mind.

therealprof (approved):
LGTM, thanks.

bors r+

therealprof (commented):
bors r-

therealprof (approved):
Now looks good to me for real.

bors r+

PR #360, opened 2021-08-15
embedded_time

burrbull:


AlyoshaVasilieva:
Speaking as a random user: if this HAL is going to change how time works, it would be nice if it fixed that `Timer`s use Hertz instead of a time unit, like embedded-hal says they should. I was working on a driver a while back that needed a timeout >1 second; I wrote it using using `core::time::Duration` which meant this HAL couldn't handle it as 1. this HAL uses a non-time unit for `Time`, and 2. this HAL's timers cannot handle >1 second countdowns (due to specifying countdowns as Hertz).

Issue #361, opened 2021-08-17
Implement PWM channels for timers 9-14


rthinman:
PWM channels `timer::PinCx` are only implemented in timer.rs for the GPIO pins used by TIM 1-5, & 8.  It would be nice to be able to use the other PWMs available in these microcontrollers.  

Separately, and perhaps this is better as another issue, but while looking into why I couldn't use TIM9 as a PWM, I noticed something else might be awry: In the latest commits on master, in pwm.rs, the match pattern for `pwm_all_channels!()` is different than in `pwm_2_channels!()` and `pwm_1_channel!()`.

```
macro_rules! pwm_all_channels {
    ($($TIMX:ident: ($timX:ident),)+) => {
...
```
Notice that $TIMX matches an identifier, whereas in the next two macro definitions, it matches a type:
```
macro_rules! pwm_2_channels {
    ($($TIMX:ty: ($timX:ident),)+) => {
...
```

```
macro_rules! pwm_1_channel {
    ($($TIMX:ty: ($timX:ident),)+) => {
...
```

Is this intentional?  I'm still a noob so can't tell the exact effect, but my suspicion would be that there might be no matches for the 1- and 2-channel versions.  I see that in the released v0.9.0 code, all these macros use "ident" not "ty".

rthinman:
Spreadsheet containing all the pins for F4xx parts, grouped in a way that they can be added to timer.rs.
[STM32F4xx_TimerPins.xlsx](https://github.com/stm32-rs/stm32f4xx-hal/files/7148598/STM32F4xx_TimerPins.xlsx)

rthinman:
PR merged, closing this issue.

PR #364, opened 2021-09-11
Add PWM channels for timers 9-14

rthinman:
Adds support for PWM outputs to timers 9-14 as mentioned in issue #361.  I think I got all the channels that are supported by the HAL (that is, channels 1-4 non-inverted; inverted pins, ETR, and BKIN are not covered).  A spreadsheet detailing the pins has been uploaded to the issue.

Eliminates the "tim8" configuration feature block in favor of keeping all feature gates based on the part number.  This lets the block that formerly contained TIM8 channels also contain channels for TIM10-14.

Uncomments two pins for TIM5 that are on the datasheet, and I don't see an errata suggesting they shouldn't be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- PWM channels for timers 9-14 [#364]
 - `ws2812_spi` example [#363]
 - `AsRef/AsMut<Rx/Tx` for Serial [#355]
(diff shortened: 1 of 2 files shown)


VersBinarii:
Hello team,
It would be fantastic if this could be merged.

therealprof (approved):
LGTM, thanks.

Re the commented out pins, it's quite possible that those weren't present in an earlier version of the PAC due to a bug in the SVD files.

bors r+

therealprof:
bors ping

PR #372, opened 2021-10-23
DynamicPin

burrbull:


burrbull:
Rebased.

I don't like this way, but I don't see alternative.

therealprof (approved):
LGTM

bors r+

PR #381, opened 2021-11-11
try to use fugit::Duraction for CountDown::Time

burrbull:
~~Not for SysTick for now~~ Use 1 microsecond as minimal timeout for SysTick `Counter`

This PR lets to use `fugit` units for timer including values bigger then 1 second.
Sampling of timer (time of 1 tick) now can be set in by FREQ generic constant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Added `Counter` with `CountDown<Time=fugit::TimerDuration>` and `atat::Clock` implementations [#381]
 - `Into<serial::Config>` for `Bps` [#387]
 - Added the missing DMA implementations for USART3 [#373]
 - `count_down` constructor for `Timer` -> `CountDownTimer` without start [#382]
@@ -23,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 [#382]: https://github.com/stm32-rs/stm32f4xx-hal/pull/382
 [#380]: https://github.com/stm32-rs/stm32f4xx-hal/pull/380
 [#374]: https://github.com/stm32-rs/stm32f4xx-hal/pull/374
+[#381]: https://github.com/stm32-rs/stm32f4xx-hal/pull/381
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -19,9 +19,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- [breaking-change] use `fugit::Duraction` for `CountDown::Time` [#381]
```
dbrgn, line 22: Typo: Duraction -> Duration (also in the PR title). Maybe add "instead of a frequency" to the line?

CHANGELOG.md
```diff
@@ -19,9 +19,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- [breaking-change] Remove all deprecated
```
dbrgn, line 23: All deprecated what? 🙂

CHANGELOG.md
burrbull: Thanks

CHANGELOG.md
burrbull: Don't pay attention. Accidentally deleted in previous PR

burrbull:
cc @andresv I've tried to implement `atat::Clock` also. Is this correspond to your vision?

andresv:
@burrbull yes I think currently this is what must be done, `embedded-hal` did not have suitable traits therefore `atat::Clock` was born.

It is a little bit inconvenient to use `atat::Clock` on embedded platforms that do not use fugit based timing (almost all HALs at the moment) because essentially you have to make a new type and impl Clock for that which probably involves reading timer counter from a specific register using unsafe.

Also last time I read that `CountDown` is maybe thrown away from `embedded-hal` v1.0, so I guess we do not have anything stable for a long time.

burrbull:
> It is a little bit inconvenient to use `atat::Clock` on embedded platforms that do not use fugit based timing (almost all HALs at the moment) because essentially you have to make a new type and impl Clock for that which probably involves reading timer counter from a specific register using unsafe.

Before port implementation to other HALs (and merge to this one), I must be sure it works as expected (I'm not sure it is so).

andresv:
Do you mean that `fugit` itself is maybe not suitable or what do you mean by that?

burrbull:
> Do you mean that `fugit` itself is maybe not suitable or what do you mean by that?

I mean I need someone test this and told me if I made mistake.
And yes, I still not sure about fugit.

burrbull:
Now it just adds new structs, so no breaking changes.

therealprof (approved):
Very nice!

bors r+

PR #388, opened 2021-11-19
Support for both embedded-hals (0.2.x and 1.0.0-alpha)

burrbull:


therealprof (commented):
Nice. This should probably be tested in CI, right?

burrbull:
cc @ryankurte Could you help with testing?

pdgilbert:
(newbie question - for my own sanity I really shouldn't be here.)
I've been trying to build some examples with the `e-hal` branch but I think I'm missing something simple. When I look at the tree almost everything is still using `embedded-hal:0.2.6` rather than `embedded-hal:1.0.0-alpha.6`.  I see `[dependencies.embedded-hal-one]`. Is there something I need to specify?

burrbull:
This impl uses 1.0.0-alpha.6 for now.
Unfortunately different alpha versions of embedded-hal are not compatible.
So if your driver uses other alpha version, this will not help you. Just use `0.2` with `compat` and wait for stabilization.

pdgilbert:
I understand it it is just using `1.0.0-alpha.6`, thus my problem with `rust-radio-sx127x`. That is ok, you should not spend a lot of time trying to support different alpha versions when the idea is to move past alpha. I am wondering how to force it to use `1.0.0-alpha.6` rather than `embedded-hal:0.2.6`?  When I build examples that use only the HAL (like `blink`) it is using `0.2.6` not `1.0.0-alpha.6` so I cannot see how to use it to test if my code works with `1.0.0`, and am not testing if `1.0.0` works.

pdgilbert:
I think I am slowly starting to understand. It is not just the *drivers* but I actually have to go through *every* package, even things like `cortex-m` and `panic-semihosting` and specify versions that use `embedded-hal:1.0.0-alpha.6`  Is that correct?

burrbull:
No you don't need to change `cortex-m`.
I tried to update `rust-radio-sx127x` for alpha6. See https://github.com/burrbull/rust-radio-sx127x/tree/alpha6
But it still blocked by `driver-pal` dependency. https://github.com/ryankurte/rust-driver-pal/pull/17

Also as I can see `driver-pal` is testing playground and has inverse optional dependencies. Now supports only linux-embedded-hal and `cp2130`. So it looks like you need to add f4xx-hal::Spi to https://docs.rs/driver-pal/0.8.0-alpha.5/driver_pal/hal/struct.HalInst.html first.

ryankurte:
> Also as I can see driver-pal is testing playground and has inverse optional dependencies. Now supports only linux-embedded-hal and cp2130. So it looks like you need to add f4xx-hal::Spi to [driver_pal::HalSpi](https://docs.rs/driver-pal/0.8.0-alpha.5/driver_pal/hal/enum.HalSpi.html) first

it is very much an experiment / WIP but, sorta the opposite... the idea was to pull out all the common bounds to simplify driver implementations, then provides opt-in helpers for `linux` and `cp2130` (`HalSpi`) fulfilling those bounds. it's a bit more complicated than it needs to be at this point, but, such is life with limited hours in the day.

to instantiate the radio driver you just need to call [`Sx127x::spi`](https://docs.rs/radio-sx127x/latest/radio_sx127x/struct.Sx127x.html#method.spi) with [types fulfilling the required bounds](https://github.com/rust-iot/rust-radio-sx127x/blob/master/src/util/main.rs#L78) (which creates a `driver-pal::Wrapper` internally). in practice (except for the current conflicting e-h issues) you shouldn't need to touch `driver-pal` to use the driver.

PR #390, opened 2021-11-20
Change rtic-tick example system clock to fix assert.

linuxtim:
The rtic-tick example asserts during clock setup because the code attempts to configure a clock of 8 MHz, which is below the lowest `SYSCLK_MIN` defined in `src/rcc/mod.rs` (12.5 MHz).  Change requested clock to 48 MHz (above the highest `SYSCLK_MIN` of 24 MHz).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - `Into<serial::Config>` for `Bps` [#387]
 - `count_down` constructor for `Timer` -> `CountDownTimer` without start [#382]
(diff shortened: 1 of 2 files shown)


burrbull:
Add link on PR to changelog:
https://github.com/stm32-rs/stm32f4xx-hal/blob/6f42b2cf6e8738b7f04bba3406d9e336ceeda2a8/CHANGELOG.md?plain=1#L14

linuxtim:
> Add link on PR to changelog

My thought was that since the fixed code was unreleased, the additional noise it would generate wouldn't justify a changelog entry. I've pushed a minimal change. Please feel free to edit the revised commit if it's unsatisfactory.

burrbull (approved):

PR #395, opened 2021-12-05
Wait for SDIO transmission to finish before polling card status

antonok-edm:
When testing with an eMMC module driven at frequencies above 12MHz, I found that `write_block` would consistently fail with either `Crc`, `DataCrcFail`, or `Timeout` errors.

Here's a sample probe output collected on the CLK and D0 lines at 24MHz. D0 has a relatively clean data block and corresponding CRC, but the very end of the transmission spends a clock cycle at ~1V (when powered by 3V3) and has large rippling peaks at nearby level transitions.

24MHz_clkdata_01_zm

As it turns out, the code in `write_block` is polling the eMMC (or SD card) status even while still transmitting on the data lines. Preventing the status commands from being sent until the transmission is complete (TXACT low) prevents the interference and allows writes to consistently succeed at any speed/bus width. It's unclear why the combination is problematic since the data transfer uses the data lines exclusively and the status command uses the CMD line exclusively. Regardless, there's no reason to ask for the card's status before transmission is complete, since it's guaranteed to still be busy receiving data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Finish SDIO data transmission before querying card status in `write_block` [#395]
 - Unify alternate pin constraints [#393]
 - [breaking-change] Use `&Clocks` instead of `Clocks` [#387]
 - Split and rename `GetBusFreq` -> `BusClock`, `BusTimerClock` [#386]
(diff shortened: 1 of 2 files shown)


burrbull:
Clippy doesn't like blocks in conditional expressions. Rewrite with `loop`, please.

burrbull:
> Clippy doesn't like blocks in conditional expressions. Rewrite with `loop`, please. And squash commits.

antonok-edm:
Fixed.

Although, for what it's worth, that exact same `while` condition is already used in several places in `sdio.rs`.

burrbull (approved):

jkristell:
@antonok-edm @burrbull It's been on my TODO list for some time to clean up the conditionals to silence clippy. I have a local branch with the changes. Gonna test the changes on my board tonight and submit it.

Issue #282, opened 2021-03-25
DMA Transfer consumes peripheral


jonnyando:
Right now there aren't really any examples of both DMA and peripherals using this HAL, so I might be doing something wrong.
I'm able to both build a DMA `Transfer` and initiate a peripheral such as the adc using this HAL, but not both at the same time.

```rust
// DMA init
let stream_0 = StreamsTuple::new(dp.DMA2).0;
let buf = singleton!(: [u16; 128] = [0; 128]).unwrap();
let dma_config = DmaConfig::default();

let adc_dma_transfer = Transfer::init(stream_0, dp.ADC1, buf, None, dma_config);  //<-- consumes dp.ADC1

// adc init
let adc_config = hal::adc::config::AdcConfig::default();
let adc_periph = Adc::adc1(dp.ADC1, true, adc_config); //<-- cannot init this peripheral. ADC1 has already been moved
```

I haven't dug too much into the `DMA` implementation but I don't think it would need ownership of the peripheral.

jonnyando:
Reading more into it, I understand I could instead hand the dma `Transfer` a copy of its address with something like
```rust 
unsafe { &*ADC1::ptr() as *const _ as u32 }
```
But wouldn't it be more elegant for that to happen inside the dma module and not in the application code?
The `stm32f1xx-hal` has dma implementations of peripherals within their respective peripheral crate, so this de/referencing happens there. Is that an approach this HAL is looking to move toward?

thalesfragoso:
The idea is to implement the necessary abstraction and traits to the Adc HAL type, and pass it to the `Transfer` API instead of the pac type. It seems that this was done already for serial, but other peripherals are still missing.

gernoteger:
I just had a look: as of now  (0.10.1) the Adc traits seem to be implemented for at least ADC1; nevertheless, the adc hal types are consumed themselves. This renders the calibration data inaccessible, as can be seen in the adc_dma example.
The peripheral in the Transfer structs is passed  as &mut to the closures of start,pause, so I don't think we get around that.

A way to fix this special issue could be to expose a reference to the peripheral. This allows us to use for example adc.sample_to_millivolts() wherever the sampled values are consumed.

PR #396, opened 2021-12-06
adc improvements for dma

gernoteger:
This PR should fix adc with dma issues as mentioned in #282: "DMA Transfer consumes peripheral".

Changes:
- DMA Transfer exposes a reference to the peripheral, currently only for PeripheralToMemory
- the ADC provides a closure similar to sample_to_millivolts
- a demonstration of its use in the adc_dma_rtic example.

As of now this PR is not complete; I kindly ask to review the concept before I complete the implementation, documentation and squashing.

### DIFF
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -23,7 +23,7 @@ use crate::{pac, rcc};
 pub mod traits;
 use traits::{
     sealed::{Bits, Sealed},
-    Channel, DMASet, Direction, Instance, PeriAddress, Stream, StreamISR,
+    Channel, DMASet, Direction, Instance, PeriAddress, SafePeripheralRead, Stream, StreamISR,
 };
 
 /// Errors.
@@ -1011,6 +1011,20 @@ where
         };
         self.next_transfer_with_common(new_buf, ptr_and_len, false, CurrentBuffer::FirstBuffer);
         Ok(r.1)
+    }
+}
+
+impl<STREAM, PERIPHERAL, BUF, const CHANNEL: u8>
+    Transfer<STREAM, PERIPHERAL, PeripheralToMemory, BUF, CHANNEL>
+where
+    STREAM: Stream,
+    ChannelX<CHANNEL>: Channel,
+    PERIPHERAL: PeriAddress + DMASet<STREAM, PeripheralToMemory, CHANNEL> + SafePeripheralRead,
+    BUF: StaticWriteBuffer<Word = <PERIPHERAL as PeriAddress>::MemSize>,
+{
+    /// Access the owned peripheral for reading
+    pub fn peripheral(&self) -> &PERIPHERAL {
+        &self.peripheral
     }
 }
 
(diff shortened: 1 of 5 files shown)


burrbull:
It is interesting way.
Is there sense to convert to `u16` implicitly if you already have `u32`?

Squash, please.

burrbull:
cc @gernoteger
Rebase and squash

gernoteger:
> It is interesting way. Is there sense to convert to `u16` implicitly if you already have `u32`?
The u16 was also used original sample_to_millivolts, I just didn't want to change maths in this PR. Thinking of it, this conversion (also in the original)  may lead to a loss of resolution:  when using 12 bit ADCs, we get 4096 possible steps from the ADC, while a 3.3V VRef will, only produce  3300 Steps.

burrbull:
You've broken git

gernoteger:
Sorry. I'm confused; should I quit working on this  PR, and continue with #406 instead?

burrbull:
> 

This PR is preferable but it should be 1 - 2 commits **rebased** on master branch with no `Merge`.

gernoteger:
I fixed the git issue. Still, I'd like to tackle the issue with the lost resolution by converting samples to milliseconds. Probably this con go to a different PR, what do you think?

src/dma/mod.rs
```diff
@@ -1059,6 +1059,11 @@ where
         }
     }
 
+    /// Access the owned peripheral for reading
+    pub fn peripheral(&self) -> &PERIPHERAL {
```
burrbull, line 1063: DMA code for this HAL is not mine, so I need to clarify is such access safe?

burrbull: cc @thalesfragoso

burrbull:
> I fixed the git issue. Still, I'd like to tackle the issue with the lost resolution by converting samples to milliseconds. Probably this con go to a different PR, what do you think?

I'm afraid I don't understand what lost resolution you mean. milliseconds? time?

gernoteger:
When converting a sample to millivolts, with a 3.3V (or less) VRef, we have
<=3300 possible steps, while a 12Bit ADC has 4096. Probably adding a
function "sample_to_microvolts" and/or direct access to the calibration
data would help here. It's related to this PR, but somewhat orthogonal.

Am Di., 4. Jan. 2022 um 19:45 Uhr schrieb Zgarbul Andrey <
***@***.***>:

gernoteger:
I documented the resolution issues mentioned earlier in #412. I think we should handle this in a different PR.
From my point of view this PR is feature complete and can be merged.

src/dma/mod.rs
burrbull: @gernoteger I still can't merge this as this method gives unsafe access any DMA payload.
Possible solution:
https://github.com/stm32-rs/stm32f4xx-hal/commit/b2752339a04cc631e37272e21bdcf5429532be25

gernoteger: thanks for doing all the work. I incorporated your suggestion nearly 1:1, only fixed a small typo in a comment

burrbull (approved):

PR #397, opened 2021-12-07
fix #362: ADC voltage conversion might be incorrect

### LINKED ISSUE #362: ADC voltage conversion might be incorrect
Hello, I would like to know if there is an error for ADC digital-to-voltage conversion. Currently, it uses `voltage = ADC_code * Vref / max_sample`, where `max_sample` is always `1 << bit_resolution - 1` instead of just `1 << bit_resolution`.

https://github.com/stm32-rs/stm32f4xx-hal/blob/979ff1555f79e3e268a600caa5c43b8d3de95217/src/adc.rs#L796-L801

This would be the contrary of most ADCs, where the divisor is **2^bits**. There's also a comment on StackExchange claiming that this should be the standard ([link](https://stackoverflow.com/a/7895054)).

It makes more sense to me that (2^bits-1) is indeed incorrect, but I'm also surprised that the STM32 official HAL seems to adopt the same approach as this Rust HAL. Relevant code here: [link](https://github.com/STMicroelectronics/STM32CubeF4/blob/4aba24d78fef03d797a82b258f37dbc84728bbb5/Drivers/STM32F4xx_HAL_Driver/Inc/stm32f4xx_ll_adc.h#L1603-L1604).

It'd be great to hear from others. Thanks!

gernoteger:
since the discussion in #362 seems to come to a conclusion, and my personal measurements supported that, I think it's time to actually fix it.
This will change max_sample from 1<<n-1 to 1<<n,
After this PR ADC readings will change, but will be more accurate than before. Users might be confused since they won't get a reading of 3300mV, but this is correct behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [breaking-change] Make `Alternate` generic over `Otype` instead of separate `Alternate` and `AlternateOD` [#383]
 - [breaking-change] Bump `stm32f4` to 0.14. Update RTIC based examples to use `rtic` 0.6 [#367]
 - [breaking-change] Bump `bxcan` to 0.6 [#371]
+- fix #362: ADC voltage conversion might be incorrect [#397]
 
+[#367]: https://github.com/stm32-rs/stm32f4xx-hal/pull/397
 [#367]: https://github.com/stm32-rs/stm32f4xx-hal/pull/367
 [#371]: https://github.com/stm32-rs/stm32f4xx-hal/pull/371
 [#383]: https://github.com/stm32-rs/stm32f4xx-hal/pull/383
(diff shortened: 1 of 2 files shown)


burrbull:
Maybe add some description about convertion rules in `adc.rs` docs?

gernoteger:
Ok, will do that

jordens (dismissed):
Thanks!
This makes the [`Debug::fmt()` output](https://github.com/stm32-rs/stm32f4xx-hal/blob/9f4d8d7a7c07c57b64fed950e22e64e44ca92596/src/adc.rs#L615-L616) a bit confusing as now that `max_sample` is exclusive. But it says nowhere that `max_sample` must be inclusive so I'm absolutely fine with it. Maybe a small comment in on the struct field would help though.

gernoteger:
@jordens Sorry, I don't understand the comment about Debug::fmt(), this PR doesn't change anything there. Sure you commented on the right PR?

gernoteger:
Thanks, now I get it. Since docs are an open task, I'll consider that.

Robert Jördens ***@***.***> schrieb am Do., 9. Dez. 2021,
18:01:

burrbull (approved):

PR #398, opened 2021-12-08
Fix clippy warnings

jkristell:
Rewrite the loop conditionals to silence clippy.

Tested on a feather-f405 board.


burrbull (approved):

jkristell:
I Think the failure on f417 is a glitch in the CI. It builds locally.

PR #399, opened 2021-12-09
Add internal_pull_down() method

Windfisch:
PR #266 has added a `internal_pull_up()` function. This PR adds `internal_pull_down()` for symmetry reasons

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Implementation of RTIC Monotonic for TIM2 & TIM5 under `rtic` feature [#380] [#390]
 - `IoPin` for `Output<OpenDrain>> <-> Input<Floating>>` [#374]
 - `IoPin` for `Output<PushPull>> <-> Input<PullUp>> and Input<PullDown>>` [#389]
+- Add `internal_pull_down` to `Pin<Output<OpenDrain>>` and `Pin<Alternate<PushPull>>` for symmetry
+  with `internal_pull_up` [#399]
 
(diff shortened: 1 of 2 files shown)


burrbull:
Change to `self` also this one then:
https://github.com/stm32-rs/stm32f4xx-hal/blob/d1123d896feabecb17148423e05a1495782484bd/src/gpio.rs#L303

Is this relevant to other modes. `Alternate<OpenDrain, A>` for example?

Windfisch:
I have done the opposite, using the `pub fn (&mut self, on: bool) -> ()` signature everywhere, instead of the pointless "taking Self, returning Self", because in no case there was a change in the type state.

One might want to think about whether type-stating this pull-up/down resistor state, but for now, I'd leave it. If we do this, we can easily change it back with only minor modifications needed in user code. Also, no feature I am aware of actually requires the pins in one particular pullup/down state, to typestating it might actually cause more harm than good.

Does that look good to you?

Windfisch:
Ignore what I wrote above. I will be doing it the other way round, using the "pointless-y self-consume and returning Self", because this enables us to do things like  https://github.com/stm32-rs/stm32f4xx-hal/blob/d1123d896feabecb17148423e05a1495782484bd/examples/ssd1306-image.rs#L39-L48

burrbull:
Merge conflict

Windfisch:
fixed (hopefully)

Windfisch:
@burrbull do you have time to review and merge this O:)?

burrbull:
> @burrbull do you have time to review and merge this O:)?

LGTM, but, please rebase and squash instead of merge.

Windfisch:
@burrbull there you go

burrbull (approved):

PR #401, opened 2021-12-25
use RTIC 1.0, fix warnings

burrbull:


therealprof (changes requested):
I think moving to Edition 2021 is maybe a bit too early. I don't see the pressing need to limit the possible Rust compilers just yet. If we really want to go there now we should do another edition 2018 release first.

burrbull:
> I think moving to Edition 2021 is maybe a bit too early. I don't see the pressing need to limit the possible Rust compilers just yet. If we really want to go there now we should do another edition 2018 release first.

Ok. Resonable. Reversed

therealprof (dismissed):
LGTM, thanks.

burrbull:
bors r=therealprof

burrbull:
bors retry

burrbull:
bors retry

burrbull:
Four tries and each time 1 of tasks freezes.

burrbull:
bors retry

burrbull:
bors r=therealprof

therealprof (approved):
bors r+

therealprof:
> Four tries and each time 1 of tasks freezes.

Seems GH is having weird issues...

PR #405, opened 2021-12-28
Correct default mode for debugger pins

burrbull:


therealprof (approved):
LGTM, thanks.

bors r+

PR #406, opened 2021-12-29
 adc improvements for dma (rebased)

burrbull:
#396

### DIFF
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -1059,6 +1059,11 @@ where
         }
     }
 
+    //### ge experiment
+    pub fn peripheral(&self) -> &PERIPHERAL {
+        &self.peripheral
+    }
+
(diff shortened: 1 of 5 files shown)


burrbull:
Not sure about `fn peripheral`

src/dma/mod.rs
```diff
@@ -1059,6 +1059,11 @@ where
         }
     }
 
+    //### ge experiment
+    pub fn peripheral(&self) -> &PERIPHERAL {
+        &self.peripheral
+    }
+
```
therealprof, line 1066: What's the point of this? Also the comment is a bit weird. 😅

burrbull: As I understand this is method on `Transfer` for accessing to peripheral payload.
But I can't say is this safe.

PR #409, opened 2021-12-30
Fix Width of TIM5 F410

burrbull:


therealprof (approved):
LGTM, thanks.

bors r+

PR #419, opened 2022-01-16
Prevent overflows optimizing SAI PLL

dgoodlad:
Setting the common target PLL freq of 49.152 MHz for the SAI clock (192kHz * 256) results in an overflow during multiplication. This patch catches the overflow, though I'm sure there's a better approach to optimizing the PLL settings that could avoid this.


burrbull:
Add to changelog, please.

burrbull (dismissed):

burrbull:
rustfmt & squash

burrbull (approved):

PR #424, opened 2022-01-24
i2c inherent

burrbull:


burrbull:
Need help on splitting Acknowledge errors

PR #429, opened 2022-02-02
Implement embedded-storage traits

chemicstry:
This PR implements [embedded-storage](https://github.com/rust-embedded-community/embedded-storage) traits for flash.

One major headache with F4 series is dual-bank flash and non-uniform sector sizes, which required quite a bit of code to abstract away. I went through all of the F4 reference manuals and believe that `flash_sectors(...)` function should be correct for all variants. AFAIK, there is no single source about flash layout for all the chips. Moreover, ST [lists](https://www.st.com/en/microcontrollers-microprocessors/stm32f4-series.html#products) F429 and F469 with 512 KB of flash as dual-bank, but there is no information about it in the reference manual. I believe this could be an error in the website and only 1 MB chips have dual-bank capabilities.

The PAC crate is also missing `DB1M` field for some of the dual-bank capable chips, so for now I hardcoded the bit position in `OPTCR` register.

The writable `NorFlash` trait was implemented for the largest sector size of 128 KB, because `embedded-storage` does not intend to supprot non-uniform sectors (see [comment](https://github.com/rust-embedded-community/embedded-storage/issues/23#issuecomment-1028112586)). Smaller sectors are erased together in the larger 128 KB group. There was a suggestion to add different types for the smaller sector ranges, but I'm not sure if that is useful?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,12 +23,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `Pwm` struct with `split` method and implementation of embedded-hal::Pwm (similar to f1xx-hal) [#425]
 - VSCode setting file
 - Add CAN1 PB8/PB9 and SPI3 MOSI PC1 pin mappings for F446 [#421]
+- Add embedded-storage traits for flash [#429]
 
 [#421]: https://github.com/stm32-rs/stm32f4xx-hal/pull/421
 [#422]: https://github.com/stm32-rs/stm32f4xx-hal/pull/422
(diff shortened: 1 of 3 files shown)


burrbull:
LGTM. Could you add example?

chemicstry:
Adding an example is problematic, because that requires modifying linker script to add an exclusive memory region for flash writing/reading and it has to not interfere with firmware binary size as well as total flash size. It would not be possible to have a single linker script that also supports 64K devices.

What if instead I added an inline example in the docs?

burrbull (approved):

PR #436, opened 2022-02-07
Use time instead of rtcc, add rtc-example

burrbull:
Ported from f7xx-hal

cc @systec-ms


therealprof (dismissed):
Interesting that using time instead of rtcc removes so much code.

bors r+

burrbull:
Rebased

therealprof (approved):
bors r+

PR #423, opened 2022-01-23
move all fugit-dependent implementations in one module

burrbull:
Add helper aliases and traits in prelude.
Add simple `pwm-sinus` example

PR #452, opened 2022-02-17
extend 16-bit Delay's to 32

burrbull:


src/timer/delay.rs
```diff
@@ -114,22 +114,28 @@ pub type DelayMs<TIM> = FDelay<TIM, 1_000>;
 
 impl<TIM: Instance, const FREQ: u32> FDelay<TIM, FREQ> {
     /// Sleep for given time
-    pub fn delay(&mut self, time: TimerDurationU32<FREQ>) -> Result<(), Error> {
```
therealprof, line 117: I'm not too excited about this undocumented breaking API change, maybe it would be better to keep the `Result` for now?

burrbull: This version of Delay was added in #423 and not released yet.
So if now it is infallible what sense to keep Result ?

therealprof: Ah, I missed that. If all of the fallible API was just recently added, that's of course fine to change it back to infallible.

burrbull:
bots retry

PR #458, opened 2022-02-24
Support eMMC in SDIO module

antonok-edm:
Followup to https://github.com/jkristell/sdio-host/pull/10, which will also be required before this PR can be merged (cc @jkristell)

This PR adds support for eMMC peripherals using the SDIO module. `Sdio` is now generic over an `SdioPeripheral`, which can either be a `SdCard` or `Emmc`. Many of the methods are generic enough to be shared between both, although some (like initialization) need to be specialized for each.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ nb = "1"
 rand_core = "0.6"
 stm32f4 = "0.14.0"
 synopsys-usb-otg = { version = "0.2.0", features = ["cortex-m"], optional = true }
-sdio-host = { version = "0.5.0", optional = true }
+sdio-host = { version = "0.6.0", optional = true }
 embedded-dma = "0.2.0"
 bare-metal = { version = "1" }
 void = { default-features = false, version = "1.0.2" }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -34,7 +34,7 @@ nb = "1"
 rand_core = "0.6"
 stm32f4 = "0.14.0"
 synopsys-usb-otg = { version = "0.2.0", features = ["cortex-m"], optional = true }
-sdio-host = { version = "0.5.0", optional = true }
+sdio-host = { version = "0.5.0", git = "https://github.com/antonok-edm/sdio-host", branch = "emmc", optional = true }
```
burrbull, line 37: What with this?

antonok-edm: https://github.com/jkristell/sdio-host/pull/10 is now merged upstream.

@jkristell - let me know when that makes it to a release on crates.io and I'll update the line here.

jkristell: sdio-host 0.6 should be on crates.io now!

antonok-edm: Awesome, updated here as well now.

burrbull (approved):

burrbull:
bors merge

burrbull:
bors merge

PR #463, opened 2022-03-13
Dual i2s

YruamaLairba:
Hi, i'm trying to implement full duplex I2s. at the moment, this branch use a modified version of stm32_i2s crate i'm also working on. the modified version is included through submodule

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "stm32_i2s"]
+	path = stm32_i2s
+	url = https://github.com/YruamaLairba/stm32_i2s
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ version = "1.0.0-alpha.7"
 package = "embedded-hal"
 
 [dependencies.stm32_i2s_v12x]
-version = "0.2.0"
+version = "0.3.0-alpha"
 optional = true
 
 [dev-dependencies]
@@ -85,6 +85,9 @@ ws2812-spi = { version = "0.4.0", features = [] }
 version = "0.3"
 default-features = false
 features = ["macros"]
+
+[patch.crates-io]
+stm32_i2s_v12x = { path = "stm32_i2s" }
 
 [features]
 device-selected = []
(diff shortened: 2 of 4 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "stm32_i2s"]
+	path = stm32_i2s
+	url = https://github.com/YruamaLairba/stm32_i2s
```
burrbull, line 3: Please, don't use submodules.
Cargo has powerful dependency control system.
```suggestion
[dependencies.stm32_i2s]
git = "https://github.com/YruamaLairba/stm32_i2s"
```
if you need to rename package (for example if you need to use 2 similar crate) add
```toml
package = "nename"
```

burrbull:
As I understand, you need also define pins for `ExtSd` which require specific alternate function:
see `gpio::alt` mode and datasheets like https://www.st.com/resource/en/datasheet/DM00037051.pdf

burrbull:
Try this:
https://github.com/stm32-rs/stm32f4xx-hal/tree/i2sext2

.gitmodules
YruamaLairba: Sorry, the reason i use submodule is to avoid to have to do a push each time i fix something in smt32_i2s while ensuring other people can build to check if it work.

YruamaLairba:
> Try this: https://github.com/stm32-rs/stm32f4xx-hal/tree/i2sext2
i was not expecting to have a review now. i just open that to signal i'm working on it.

And sorry, what you did is wrong, you assumed I2sEXT exist as long as SPi2 and SPI3 exist, but stm32f410 and stm32f446 doesn't have extended I2S.

burrbull:
> > Try this: https://github.com/stm32-rs/stm32f4xx-hal/tree/i2sext2
> 
> i was not expecting to have a review now. i just open that to signal i'm working on it.
> 
> And sorry, what you did is wrong, you assumed I2sEXT exist as long as SPi2 and SPI3 exist, but stm32f410 and stm32f446 doesn't have extended I2S.

This is just to give you a hint, not for replacing your PR. I'm not familiar with I2s so I can't do PR for you.
Possibly @samcrow could help.

Issue #445, opened 2022-02-12
transactional i2c


burrbull:
Although #443 was merged new Transactional API are still not implemented:
https://github.com/stm32-rs/stm32f4xx-hal/blob/aa934ca437c18a682dbcb8a7550ed8dffb9a7298/src/i2c/hal_1.rs#L65
https://github.com/stm32-rs/stm32f4xx-hal/blob/aa934ca437c18a682dbcb8a7550ed8dffb9a7298/src/spi/hal_1.rs#L172

burrbull:
@ryankurte as you are interested in this hal could you help with this issue?

PR #464, opened 2022-03-13
Added transactional SPI implementation

ryankurte:
per #445, i don't have a moment to look at the I2C at the moment as the behaviour is a bit more complex wrt. where starts and ends get asserted.


burrbull:
Update changelog, please.

burrbull (approved):

burrbull:
Thank you.

bors r+

PR #465, opened 2022-03-15
Fix defmt support

jkristell:
Right now on relase 0.12 (and master) this:
```
DEFMT_LOG=debug cargo build --release --features=rt,stm32f405,defmt --examples
```
fails to compile due to `defmt` support in fugit not being enabled.

This workaround fixes that, but feels like a hack.
It's described here: https://blog.turbo.fish/cargo-features/

The other option is to rename the fugit feature under a different name, but this fix could be used until then.

Let me know if you want this to be merged and I can write a Changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Move pin mode at the end of generics, add defaults for modes,
   bump MSRV to 1.59 [#418]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


burrbull:
LGTM. Thanks

burrbull:
Could you also move `hd44780-driver` to `dev-dependencies`?

jkristell:
Changelog updated and I added a commit with the hd44780-driver moved to dev-dependencies.

burrbull (approved):

Issue #483, opened 2022-04-18
SPI Slave mode


Arnaud078:
Hello,

I am working on a project which requires a Raspberry Pi to communicate with a stm32f411. I was thinking of using SPI with the Raspberry as the master. As of now, neither this crate nor the embedded-hal seem to provide this feature. I have forked this repo and implemented a working solution here : https://github.com/Arnaud078/stm32f4xx-hal/commit/4b4baa5ebee1a5b8b0553fbe93ad7343efe0b4d0

I am new to rust, open source and this repo, so I am wondering if that looks good enough for a PR or is there more work to be done ? Is there a cleaner way to achieve SPI slave mode ?

Rgds,
Arnaud

burrbull:
If you have ideas with working code you are free to make PR.
It's better to discuss it there.
In first view your code looks good.

burrbull:
https://github.com/stm32-rs/stm32f1xx-hal/pull/408

Arnaud078:
solved by https://github.com/stm32-rs/stm32f4xx-hal/pull/487

PR #487, opened 2022-04-23
adding spi slave

Arnaud078:
as mentionned in issue https://github.com/stm32-rs/stm32f4xx-hal/issues/483

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Changed
+
+- `Spi` can be operated as `Slave` [#487]
+
+[#487]: https://github.com/stm32-rs/stm32f4xx-hal/pull/487
+
 ## [v0.13.1] - 2022-04-20
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


Arnaud078:
I tested that roundtripping between all the modes is fine with :
let spi3 = Spi::new_slave(spi, (sck, miso, mosi), mode, 8_000_000.Hz(), &clocks);
        let spi3 = spi3.to_master_operation();
        let spi3 = spi3.to_bidi_transfer_mode();
        let spi3 = spi3.to_slave_operation();
        let spi3 = spi3.to_normal_transfer_mode();

Slave mode is able to transmit.

I can't test that nothing was broken in bidi or master mode though.

burrbull:
Could you add any tested example?

Arnaud078:
added tested example of spi slave with dma in rtic

burrbull:
Add example to Cargo.toml with `rtic` feature.

What about `embedded-hal` traits? Are they usefull for master mode only?

burrbull:
rustfmt

Arnaud078:
Added example to Cargo.toml. I am not sure it will solve all the build errors though, I am using undeclared crates :/

I checked embedded-hal, but I don't see any reference to slave or master mode. I don't think I can reference other traits than what was already referenced. I think we can add the 2 new type structs there (Master & Slave), as it is now duplicated in stm32f1xx and stm32f4xx.

burrbull (approved):

Arnaud078:
Great ! Thank you for your time.

Issue #505, opened 2022-07-20
Releasing drivers / pins


schteve:
HAL drivers naturally need to own some peripherals while they are in use. However, it should be possible to safely release the peripheral ownership (destroying the driver in the process) so the peripherals can be used for something else. This is necessary if you want to use the same pin for several purposes at different times.

Some drivers already seem to implement this, at least in part. For example there is a Serial::release(), however this only works if you have the serial struct still (calling split() on it separates the Tx and Rx and prevents being able to release them afterwards). In this example it seems like the most natural thing to do would be to 'unsplit' Tx and Rx back into a Serial and then release Serial. But, I don't see anything to that effect.

Is there already an existing way to handle this? If not, does it make sense to add one?

To illustrate another example, you can run PWM on a timer channel, but you might want to reclaim that timer and pin to use for other purposes. I'm imagining it would look something like below. Note there is already a PwmHz::release(), but at the point I would like to release it I only have a PwmChannel.

```rust
if let (Some(dp), Some(cp)) = (
    stm32f4xx_hal::pac::Peripherals::take(),
    cortex_m::peripheral::Peripherals::take(),
) {
    let rcc = dp.RCC.constrain();
    let clocks = rcc.cfgr.freeze();

    let gpioa = dp.GPIOA.split();

    let mut pwm_ch1 = dp.TIM2
        .pwm_hz(gpioa.pa5.into_alternate(), 20.kHz(), &clocks)
        .split();
    pwm_ch1.set_duty(50);
    pwm_ch1.enable();

    // Do other things. Later...
    // Release PA5 and TIM2
    // let (pa5, tim2) = pwm_ch1.release();
    
    // Use PA5 and TIM2 for other purposes
    // ...
}
```

burrbull:
> Note there is already a PwmHz::release(), but at the point I would like to release it I only have a PwmChannel.

`PwmChannel` erases information about Pin, so it is impossible to restore it (same as `Rx` and `Tx` in serial). The only way is to keep `Pin` inside `PwmChannel` but not everyone agrees with this as it makes type more complicate.

schteve:
[...]

PR #514, opened 2022-08-15
Serial Tx, Rx containing pins

burrbull:
Related to #505 

Instead of changing API I added alternative.

So now to can choose between `Tx`, `Rx` with erased information about pins and `UTx`, `URx` which can join to restore `Serial`

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -332,6 +332,123 @@ impl<USART, PINS, WORD> AsMut<Rx<USART, WORD>> for Serial<USART, PINS, WORD> {
     }
 }
 
+/// Serial receiver containing RX pin
+pub struct URx<USART, RX, WORD = u8> {
+    _usart: PhantomData<USART>,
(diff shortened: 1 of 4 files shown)


src/serial.rs
```diff
@@ -614,6 +731,12 @@ impl<USART: Instance, PINS, WORD> Serial<USART, PINS, WORD> {
     }
 }
 
+impl<USART: Instance, TX, RX, WORD> Serial<USART, (TX, RX), WORD> {
+    pub fn split_nondestructive(self) -> (UTx<USART, TX, WORD>, URx<USART, RX, WORD>) {
+        (UTx::new(self.usart, self.pins.0), URx::new(self.pins.1))
+    }
```
burrbull, line 737: should be better name

therealprof (approved):
LGTM, thanks.

bors r+

burrbull:
bors retry

PR #517, opened 2022-08-25
Apply #[inline] attribute to bitbanding functions.

therealprof:
The generated code for the functions is usually only 1-2 instruction(s),
calling such a function causes more overhead than inlining it. Most
example binaries shrink up to 1kB by this hint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+ - apply #[inline] attribute to bitbanding functions [#517]
  - update `stm32f4` to 0.15.1 [#481]
  - use `stm32_i2s_v12x` version 0.3, reexport it, and implements requirement for it [#490]
  - i2s module don't reuse marker from spi module and define its own [#490]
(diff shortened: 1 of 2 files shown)


burrbull (approved):

burrbull:
should we use "always" to make it inline in debug mode?

burrbull:
what about those `assert`s
https://github.com/stm32-rs/stm32f4xx-hal/blob/fce6276ceb23e2d52c8ea896bcb2c414f702e8e1/src/bb.rs#L45-L46

maybe use `debug_assert` instead?

therealprof:
> should we use "always" to make it inline in debug mode?

I wouldn't. There's quite some code in there and I assume that would bloat massively in debug mode.

therealprof:
> maybe use `debug_assert` instead?

Not sure, it might have an contrary effect, i.e. the compiler not seing the constraints anymore causing more code to be generated. 

Just for reference, this is the generated assembly:
```
stm32f4xx_hal::bb::write:
.Lfunc_begin5:
	.fnstart
	.cfi_startproc
	movw r1, #512
	movt r1, #16967
.Ltmp93:
	str r0, [r1]
.Ltmp94:
	bx lr
```

PR #518, opened 2022-08-26
fix PwmHz::get_period

burrbull:
See https://github.com/stm32-rs/stm32f1xx-hal/pull/434


therealprof (approved):
LGTM, thanks.

bors r+

PR #521, opened 2022-08-27
const BIDI: bool

burrbull:


src/spi.rs
```diff
@@ -585,13 +505,13 @@ impl<SPI: Instance, PINS, TRANSFER_MODE, OPERATION> Spi<SPI, PINS, TRANSFER_MODE
 
     /// Return `true` if the TXE flag is set, i.e. new data to transmit
     /// can be written to the SPI.
-    pub fn is_txe(&self) -> bool {
+    pub fn is_tx_empty(&self) -> bool {
```
therealprof, line 508: What is the motivation for these renames?

burrbull: This is not principal.
Main motivation is compatibility with f1xx-hal:
https://github.com/stm32-rs/stm32f1xx-hal/blob/6c6c8c5154510b519fdf74207d5ff626accc0e39/src/spi.rs#L450

burrbull: Revert? Or make copy and deprecate?

therealprof: I hadn't thought of that, but copy and deprecate sounds better to me.

burrbull: done. See last commit

therealprof (approved):
LGTM, thanks.

bors r+

burrbull:
bors retry

PR #523, opened 2022-08-27
Spi u16 frame size & optimized SPI write

burrbull:
Depends on #521


burrbull:
cc @adamgreig @therealprof

src/spi.rs
```diff
@@ -676,6 +689,42 @@ impl<SPI: Instance, PINS, const BIDI: bool, W: FrameSize, OPERATION>
...
+                        return Err(Error::ModeFault);
+                    }
+                    break;
+                }
+            }
+        }
+        // Wait for final TXE
+        while !self.is_tx_empty() {}
+        // Wait for final !BSY
+        while self.is_busy() {}
```
burrbull, line 720: This is 1 moment I'm not sure. Do we need to check BSY here?
It takes 3-4 SPI clocks.

burrbull: I moved `busy` check to `Spi::disable`

PR #524, opened 2022-08-28
Spi u16 frame

burrbull:
part of #523


therealprof (approved):
LGTM, thanks.

bors r+

PR #529, opened 2022-09-09
Revised ADC temperature sensor pins

lborg019:
For each MCU listed checked ST's reference manual for temperature sensor ADC input pin.
Some are exclusively `ADC1_18`, others have `ADC1_16` available, and some are shared. Opted to use `ADC1_16` when available, always giving preference to what the manual says to use in order to read temperature.

Example:
[STM32F411 Reference Manual page 226](https://www.st.com/resource/en/reference_manual/dm00119316-stm32f411xc-e-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+ - Revised temperature sensor input pins for all MCUs #529
  - Support `u16` read/write for SPI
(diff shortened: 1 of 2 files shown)


burrbull:
Update changelog, please

burrbull (approved):

burrbull:
It looks like changes for f401/f411 not corresponds to RM and datasheets.
cc @therealprof

PR #531, opened 2022-09-19
Improve docs for Delay

dbrgn:
See diff.

screenshot-20220919-234000

### DIFF
--- a/src/timer/delay.rs
+++ b/src/timer/delay.rs
@@ -58,7 +58,32 @@ impl SysDelay {
     }
 }
 
-/// Periodic non-blocking timer that imlements [embedded_hal::blocking::delay] traits
+/// Periodic non-blocking timer that imlements [embedded_hal::blocking::delay] traits.
+///
+/// ### Example: Millisecond precision
+///
+/// To instantiate this with the TIM2 timer and millisecond precision:
+///
+/// ```rust
+/// let mut delay = dp.TIM2.delay_ms(&clocks);
+/// delay.delay_ms(320_u32);
+/// ```
+///
+/// With millisecond precision, you can wait from 2 ms to 49 days.
+///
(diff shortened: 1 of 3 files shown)


dbrgn:
Do I really have to update the CHANGELOG for a minor documentation update? Is there a way to skip this check?

src/timer/delay.rs
```diff
@@ -58,7 +58,29 @@ impl SysDelay {
...
+///
+/// With millisecond precision, you can wait from 2 ms to 49 days.
+///
+/// ### Example: Microsecond precision
+///
+/// To instantiate this with the TIM5 timer and microsecond precision:
+///
+/// ```rust
+/// let delay = dp.TIM5.delay_us(&clocks);
+/// delay.delay_us(30_u32);
```
burrbull, line 80:
```suggestion
/// let delay = dp.TIM5.delay_us(&clocks);
/// delay.delay_us(30_u32);
/// delay.delay_ms(5_u32);
/// delay.delay(5.millis());
/// delay.delay(3.secs());
```
From your docs someone can infer that only `delay_us` works with microsecond timer. This is not true, see suggestion.

Also you can initialize timer with custom precision to change minimum/maximum precision, not only 1 microsecond or 1 millisecond per clock.
For example 2kHz (500 us/clock) timer can wait from 1 ms to 24 days.
```rust
/// let delay = dp.TIM5.delay::<2_000>(&clocks);
/// delay.delay_ms(1_u32);
/// delay.delay_ms(320_u32);
```

burrbull (approved):

burrbull:
I've added `no changelog` label, but looks like it doesn't work.
Make change in changelog, please. (There are several broken links if you don't want to show this PR)

burrbull (approved):

PR #534, opened 2022-09-30
Non-Blocking I2C based on DMA

SpeedCrash100:
Adds method `use_dma` on I2c structure for creating a new structure `I2CMasterDma`. The new structure allows to use unsafe methods ended with `_dma` to send/receive data in non blocking mode and call specified `callback` at the end of transfer. The `I2CMasterDma` also have methods for blocking transfers but they returns `nb::Result` and `WouldBlock` if there are operation in progress. Clients must call `handle_dma_interrupt` in DMAx_STREAMy interrupt and `handle_error_interrupt` in I2Cx_ER interrupt for correct working of this structure.

At the current stage Start generation and Address sending is blocking anyway, so current implementation have benefits only with high amount of data to transfer. For example, sending image to display.

I've also prepared example of usage with SSD1306 display and LM75B temperature sensor in [separate repository](https://github.com/SpeedCrash100/stm32f4xx-dma-i2c-example)


burrbull:
This is great work.
Example also looks interesting (although I don't see sense in [I2c1HandleProtected](https://github.com/SpeedCrash100/stm32f4xx-dma-i2c-example/blob/1375059b5e19b90f4ade5d948f61c510c44104a0/src/main.rs#L54). Why you need extra mutex on `Shared` RTIC resource?)

cc @therealprof

SpeedCrash100:
I don't know how to pass rtic Mutex to any other structure. In example, `I2c1HandleProtected` is used in connector to display so I have to somehow pass i2c bus to it.

SpeedCrash100:
I've found a problem with current blocking API. There are instruction how software should handle NACK error in Reference Manual in section: _Inter-integrated circuit (I2C) interface -> I2C functional description -> Error conditions_ (RM0383, page 483):

> a transmitter which receives a NACK must reset the communication: ... If Master: a Stop or repeated Start condition must be generated by software

It is not a problem without non-blocking API because Start will be generated without any checks. However, with non-blocking I2C access it is required to check i2c busyness via SR2 -> BUSY bit. This bit will not cleared until Stop or Start generated as well as MSL bit. I've created a 2 commits to address this problem in Non-blocking API but a `I2CMasterDma` also provides access to blocking api by using current crate implementation. 

What should I do? Add Stop generation in blocking API when NACK occurs _or_ make changes locally for `I2CMasterDma` struct (check exit result of [`write`](https://github.com/stm32-rs/stm32f4xx-hal/blob/eec5548262802200bb6f2a909f74b07a7f0f3235/src/i2c.rs#L444), [`read`](https://github.com/stm32-rs/stm32f4xx-hal/blob/eec5548262802200bb6f2a909f74b07a7f0f3235/src/i2c.rs#L388), [`write_read`](https://github.com/stm32-rs/stm32f4xx-hal/blob/eec5548262802200bb6f2a909f74b07a7f0f3235/src/i2c.rs#L473) of `I2c` and generate stop when necessary) ?

src/i2c/dma.rs
```diff
@@ -0,0 +1,624 @@
...
+/// Callback type to notify user code of completion I2C transfers
+pub type I2cCompleteCallback = fn(Result<(), Error>);
+
+pub trait I2CMasterWriteDMA {
+    /// This function is unsafe because user must ensure that `bytes` will live until `callback` called
+    unsafe fn write_dma(
+        &mut self,
+        addr: u8,
+        bytes: &[u8],
+        callback: Option<I2cCompleteCallback>,
```
burrbull, line 27: Is `Option` needed here.
As I understand `|_| {}` is equivalent to `None` here.

SpeedCrash100:
I see a problem. The example will be built only for stm32f411 feature because DMA streams differs from device to device. Also I will push some basic docs and changes for clippy now

burrbull:
Is PR ready?

Do you have example of callback usage?

SpeedCrash100:
> Is PR ready?
> 
> Do you have example of callback usage?

PR is ready, I cannot find more problems at least.

Example of callback usage provided in example [stopwatch-with-ssd1306-and-interrupts-and-dma-i2c.rs](https://github.com/stm32-rs/stm32f4xx-hal/pull/534/files#diff-4011472252b94ecaec5ec101e31a1a1a0ce2b1216ea45a8344612a5afeada836): A two variables [`COMMAND_SEND` and `DRAWING`](https://github.com/stm32-rs/stm32f4xx-hal/pull/534/files#diff-4011472252b94ecaec5ec101e31a1a1a0ce2b1216ea45a8344612a5afeada836R84-R85) used to protect from ssd1306 driver overwrite command and data buffer while these buffers is sending. Callback usage is [here](https://github.com/stm32-rs/stm32f4xx-hal/pull/534/files#diff-4011472252b94ecaec5ec101e31a1a1a0ce2b1216ea45a8344612a5afeada836R146)

burrbull (approved):

PR #535, opened 2022-10-07
Bump synopsys-usb-otg to 0.3.1

dgoodlad:
Fixes an issue with FIFO allocation limits. tl;dr the previous version assumed that 16 32-bit words were allocated for every _possible_ TX endpoint, when only active endpoints should be considered.

See stm32-rs/synopsys-usb-otg#26


burrbull (approved):

dgoodlad:
Added an entry to the changeling in ddb9dd5

PR #539, opened 2022-10-13
Rework SPI DMA Rx/Tx: keep Spi object inside, add release method to reuse it

imihajlow:
To support txrx(), introduce two new structs, RxCoupled and TxCoupled, which work independently with DMA, but require each other to release the Spi object.


burrbull:
replaced with #616

PR #541, opened 2022-10-18
fix removed Tx::Write<u8>

### LINKED ISSUE #528: u8 .write method or Write trait missing
(Using git stm32f4xx-hal#b2636deccd331a52114a228e1ae17d51d25c016e)
I am having trouble compiling some examples,  I think because the `.write` method or `Write` trait is not found for `u8`. In a very stripped down example (below) with setup code in the main function I get 
```
55 |            block!(tx1.write((*byte))).ok();
   |                       ----- ^^^^^^^ expected `u16`, found `u8`
   |                       |
   |                       arguments to this function are incorrect
```
A `u8` method is not found but it does work by converting to `u16` with
```
block!(tx1.write(((*byte)).into())).ok();
```
The real problem is that I want to use a `setup()` function in the code, and that gives
```
error[E0624]: associated function `write` is private
   --> examples/zzz.rs:55:23
    |
55  |            block!(tx1.write(*byte)).ok();
    |                       ^^^^^ private associated function
    |
   ::: /home/paul/.cargo/git/checkouts/stm32f4xx-hal-fe8350cc04cacf3f/b2636de/src/serial.rs:990:5
    |
990 |     fn write(&mut self, word: u8) -> nb::Result<(), Error> {
    |     ------------------------------------------------------ private associated function defined here
    |
    = help: items from traits can only be used if the trait is in scope
help: the following traits are implemented but not in scope; perhaps add a `use` for one of them:
    |
5   | use core::hash::Hasher;
    |
5   | use embedded_hal::i2c::blocking::I2c;
    |
5   | use embedded_hal::serial::blocking::Write;
    |
5   | use embedded_hal::serial::nb::Write;
    |
      and 9 other candidates

```
None of the `use` suggestions seem to help.  With several other HALs this example compiles using `nb::block` but using that in place of  `stm32f4xx_hal::block` does not fix the problem. I believe this is a fairly new problem in `stm32f4xx-hal` since the example worked previously.

<details>
<summary>Click to expand example code</summary>
#![deny(unsafe_code)]
#![no_main]
#![no_std]

use panic_halt as _;

use cortex_m_rt::entry;

//use nb::block;   // this is used  with other HALs

use stm32f4xx_hal::{
    block,
    pac::Peripherals,
    pac::USART1,
    prelude::*,
    serial::{config::Config, Rx, Serial, Tx},
};

fn setup() -> (Tx<USART1>, Rx<USART1>) {
    let p = Peripherals::take().unwrap();
    let rcc = p.RCC.constrain();
    let clocks = rcc.cfgr.freeze();
    let gpioa = p.GPIOA.split();
    //p.USART1.cr1.modify(|_, w| w.rxneie().set_bit()); //need RX interrupt?
    Serial::new(
        p.USART1,
        (gpioa.pa9.into_alternate(), gpioa.pa10.into_alternate()),
        Config::default().baudrate(9600.bps()),
        &clocks,
    )
    .unwrap()
    .split()
}

#[entry]
fn main() -> ! {

    //let p = Peripherals::take().unwrap();
    //let rcc = p.RCC.constrain();
    //let clocks = rcc.cfgr.freeze();
    //let gpioa = p.GPIOA.split();
    //let (mut tx1, _rx1) = Serial::new(
    //    p.USART1,
    //    (gpioa.pa9.into_alternate(), gpioa.pa10.into_alternate()),
    //    Config::default().baudrate(9600.bps()),
    //    &clocks,
    //).unwrap().split();

    let (mut tx1, _rx1) = setup();
     

    loop {
       for byte in b"abc\r\n" {
           block!(tx1.write(*byte)).ok();
           //block!(tx1.write((*byte).into())).ok();
       }
    }
}

</details>

burrbull:
Fix #528 .
Restore implementation accidentally removed in #514

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Serial Tx, Rx containing pins [#514] [#515]
+- Serial Tx, Rx containing pins [#514] [#515] [#540]
 - Implementation of From trait for Pin-to-PartiallyErasedPin [#507]
 - Implementation of From trait for Pin-to-ErasedPin [#507]
 - Implementation of From trait for PartiallyErasedPin-to-ErasedPin [#507]
@@ -58,6 +58,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 [#529]: https://github.com/stm32-rs/stm32f4xx-hal/pull/529
 [#536]: https://github.com/stm32-rs/stm32f4xx-hal/pull/536
 [#534]: https://github.com/stm32-rs/stm32f4xx-hal/pull/529
+[#540]: https://github.com/stm32-rs/stm32f4xx-hal/pull/540
 
 ## [v0.13.2] - 2022-05-16
 
--- a/src/serial/hal_02.rs
+++ b/src/serial/hal_02.rs
@@ -51,6 +51,17 @@ mod nb {
         }
     }
 
+    impl<USART: Instance> Write<u8> for Tx<USART, u8> {
+        type Error = Error;
+
+        fn write(&mut self, word: u8) -> nb::Result<(), Self::Error> {
+            self.write(word)
+        }
+        fn flush(&mut self) -> nb::Result<(), Self::Error> {
+            self.flush()
+        }
+    }
+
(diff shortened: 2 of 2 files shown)


therealprof (approved):
LGTM, thanks.

bors r+

PR #542, opened 2022-10-18
Transactional I2C

### LINKED ISSUE #445: transactional i2c

SpeedCrash100:
A transactional HAL API implementation for I2C. Closes #445.

I have to rework i2c code to implement this:
  * `prepare_read`, `prepare_write` - methods to send start and address only
  * `write_wo_prepare`, `read_wo_prepare` - methods  like old `read`, `write` but they do not send start and address.
  * `read_bytes`, `write_bytes` - this methods now only for send/receive. The do not send start, stop or address.
  * All public api's behaviour is untouched so there are no breaking changes for users.

Note: I have to create duplicate methods `transaction_slice` and `transaction` because `Operation` doesn't implement Clone to send slice as iterator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - example of using i2s in out with rtic and interrupt.
 - example of using USB CDC with interrupts.
 - Added non-blocking I2C based on DMA [#534]
+- Added Transactional I2C API [#542]
 
 [#481]: https://github.com/stm32-rs/stm32f4xx-hal/pull/481
 [#489]: https://github.com/stm32-rs/stm32f4xx-hal/pull/489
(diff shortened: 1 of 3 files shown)


SpeedCrash100:
> LGTM. Is it tested?

I have tested on operations sequences Write-Write and Write-Read, and single Write. No problems detected in these cases.

burrbull (approved):

PR #545, opened 2022-10-20
DMA: add a method to wait for transfer completion

imihajlow:


burrbull:
Looks safe, but I'm not sure.
cc @therealprof

burrbull (approved):

PR #553, opened 2022-11-18
Add rtic-usb-cdc-echo example

no111u3:


examples/rtic-usb-cdc-echo.rs
```diff
@@ -0,0 +1,131 @@
...
+            },
+            Local { led },
+            init::Monotonics(mono),
+        )
+    }
+
+    #[idle]
+    fn idle(_: idle::Context) -> ! {
+        loop {}
+    }
```
dotcypress, line 95:
```suggestion
```

dotcypress: We can safely remove this part, cause RTIC uses wait loop by default.

burrbull (approved):

Cargo.toml
```diff
@@ -474,6 +474,10 @@ required-features = ["stm32f411", "rtic"] # stm32f411
 name = "rtic-usart-shell-ssd1306"
 required-features = ["stm32f411", "rtic"] # stm32f411
 
+[[example]]
+name = "rtic-usb-cdc-echo"
+required-features = ["stm32f411", "rtic"] # stm32f411
```
alexxy, line 479: add  "otg-fs", "usb_fs" to features

PR #555, opened 2022-12-06
Reset timer CNT register when PWM period is changed

scd31:
When changing the period of a PWM timer repeatedly, it will sporadically stop working. This is because the CNT register isn't being reset.

Minimum reproducible example:

```rust
#![no_main]
#![no_std]

// use panic_halt as _;
use panic_semihosting as _;

use cortex_m_rt::entry;
use stm32f4xx_hal as hal;

use crate::hal::{pac, prelude::*, timer::Channel};

#[entry]
fn main() -> ! {
    let dp = pac::Peripherals::take().unwrap();

    let buzzer = dp.GPIOA.split().pa0.into_alternate();

    // Set up the system clock. We want to run at 48MHz for this one.
    let rcc = dp.RCC.constrain();
    let clocks = rcc.cfgr.use_hse(25.MHz()).sysclk(48.MHz()).freeze();

    let mut pwm = dp.TIM5.pwm_hz(buzzer, 1000.Hz(), &clocks);
    pwm.set_duty(Channel::C1, pwm.get_max_duty() / 2);
    pwm.enable(Channel::C1);

    let mut delay = dp.TIM1.delay_us(&clocks);

    loop {
        pwm.set_period(1000.Hz());
        pwm.set_duty(Channel::C1, pwm.get_max_duty() / 2);
        pwm.configure(&clocks);

        delay.delay_ms(1000_u32);

        pwm.set_period(2000.Hz());
        pwm.set_duty(Channel::C1, pwm.get_max_duty() / 2);
        pwm.configure(&clocks);

        delay.delay(1.secs());
    }
}
```

Before this PR, this would work for a few cycles before the signal disappears. It would occasionally reappear every few minutes. After this PR, the code is consistently outputting a signal that alternates between 1 and 2 KHz.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
  - use register.modify instead of register.write to start PWM [#501]
  - add missing generic param for Spi::release implementation.
  - build rtic-usb-cdc-echo example [#554]
+ - reset timer cnt register when changing pwm period [#555]
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


burrbull (approved):

burrbull:
Should we also clear interrupt flags there?

scd31:
@burrbull I am not sure if we want to clear the interrupt flags. I would imagine we'd want to keep them (if an interrupt is raised right before set_period is called we probably still want to handle that interrupt)

burrbull (approved):

PR #561, opened 2022-12-13
add features to readme

burrbull:


therealprof (approved):
LGTM, please feel free to merge such trivial non-functional changes directly without requesting a review. ;)

Not sure how the HTML code will actually render since the GH preview displays them verbatim, but we'll find out.

bors r+

PR #565, opened 2022-12-28
enable GitHub Dependabot

rursprung:
this ensures that the dependencies are kept up to date. see [the docs][] for further information.

note: changelog intentionally left away as this only touches the build infrastructure and has no direct impact on consumers of the library.

[the docs]: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"


rursprung:
@burrbull: i still think that you should re-consider this, as it'd automate away the need to create PRs like #717 manually (same e.g. goes for the version updates to GitHub actions which you did as part of the v0.18 release a short while ago).

burrbull:
We need more active maintainers. Do you want to be one of them?

burrbull:
We need more active maintainers. Do you want to be one of them?

rursprung:
> We need more active maintainers. Do you want to be one of them?

i don't even have am STM32 lying around anymore 😕, so at this point: sorry, no.
my point was: having dependabot enabled actually makes life easier for you as there are _less_ things to do as a maintainer, not _more_. you can always opt to ignore a version update (just close or ignore the PR) if you don't have time.

Issue #564, opened 2022-12-26
Bug for SPI read: wrong last bit in every byte


qwerty19106:
I use `stm32f4xx-hal=0.14.0`, `cortex-m-rtic = 1.1`, tested on `STM32F411CE` and `STM32F446RE`.

Simple SPI initialization spawn this bug at the some SYSCLK and PCLK frequencies.
For example, `STM32F411CE` spawn bug at `SYSCLK=100.MHz()`, but at `SYSCLK=80.MHz()` no errors.
```
        let spi1 = dp.SPI1.spi(
            (gpioa.pa5, gpioa.pa6, gpioa.pa7),
            hal::spi::MODE_3,
            2.MHz(),
            &clocks,
        );
```

Presumably it occurs due to slow GPIO speed. This code fix bit error in my case:
```
        let mut sck1 = gpioa.pa5.into_push_pull_output();
        let mut miso1 = gpioa.pa6.into_push_pull_output();
        let mut mosi1 = gpioa.pa7.into_push_pull_output();
        sck1.set_speed(Speed::VeryHigh);
        miso1.set_speed(Speed::VeryHigh);
        mosi1.set_speed(Speed::VeryHigh);

        let spi1 = dp.SPI1.spi(
            (sck1.into_input(), miso1.into_input(), mosi1.into_input()),
            hal::spi::MODE_3,
            2.MHz(),
            &clocks,
        );
```

See [issue](https://github.com/embassy-rs/embassy/issues/1095) with the same error in other HAL for details.

I want create PR, but I need help.
The `set_speed` is method of `Pin`, but `Spi::new` get `Pins<SPI>`. I propose add `pub trait SetSpeed` to add it into `Pins<SPI>` implementation.
Are there other ways to fix this?

burrbull:
Correct pin mode for SPI and other peripherals is `Alternate<N>`. I've added possibility to pass pins to peripheral in any mode, but it was not very good idea. Possibly I'll delete it in future.
`set/set_speed` is implemented for alternate pins. So your code should looks like:
```rust
        let sck1 = gpioa.pa5.into_alternate().speed(Speed::VeryHigh);
        let miso1 = gpioa.pa6.into_alternate().speed(Speed::VeryHigh);
        let mosi1 = gpioa.pa7.into_alternate().speed(Speed::VeryHigh);

        let spi1 = dp.SPI1.spi(
            (sck1, miso1, mosi1),
            hal::spi::MODE_3,
[...]

PR #576, opened 2023-02-16
Add GPIO speed into SPI example, improve docs.

qwerty19106:
Fix 'wrong last bit in every received byte' problem, see [#564](https://github.com/stm32-rs/stm32f4xx-hal/issues/564) for details.

### DIFF
--- a/examples/spi-dma.rs
+++ b/examples/spi-dma.rs
@@ -39,8 +39,11 @@ fn main() -> ! {
         let stream = steams.4;
 
         let gpiob = dp.GPIOB.split();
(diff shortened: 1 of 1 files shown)


qwerty19106:
@burrbull 
`Spi::new` has not docs, and the `pins` param consumes Pin with any mode.
I can add `Spi::new` docs with inline example, or you will change `Spi::new` API?

burrbull:
any doc improvements are appreciated

burrbull:
Fix errors, please.

burrbull (approved):

Issue #586, opened 2023-03-14
Fix Transfer::clear_idle_interrupt trait bounds


qwerty19106:
1) Any `Transfer<..., PeripheralToMemory>` methods use this trait bounds:
```rust
BUF: WriteBuffer<Word = <PERIPHERAL as PeriAddress>::MemSize>,
```
But `clear_idle_interrupt` use `ReadBuffer`. I assume it is bug.

2) I guess `BUF` trait bounds doesn't need. I would write this code like this:
```
impl<STREAM, const CHANNEL: u8, PERIPHERAL, BUF>
    Transfer<STREAM, CHANNEL, PERIPHERAL, PeripheralToMemory, BUF>
where
    STREAM: Stream,
    PERIPHERAL: PeriAddress + RxISR,
{
    /// Clear idle line interrupt flag
    pub fn clear_idle_interrupt(&self) {
        self.peripheral.clear_idle_interrupt();
    }
}
```

3) Why [SafePeripheralRead](https://docs.rs/stm32f4xx-hal/latest/stm32f4xx_hal/dma/traits/trait.SafePeripheralRead.html#) is implemented for `Adc` only?
The `serial::Rx` has only 3 methods (`serial::RxISR` trait) which got `&self`. If we implement `SafePeripheralRead` for `serial::Rx`, then we can call `transfer.peripheral().clear_idle_interrupt()`.

If you agree with me, I can make PR. @burrbull

qwerty19106:
Fixed in #588

PR #588, opened 2023-03-14
Allow dma::Transfer.peripheral() for serial::Rx

qwerty19106:
* Impl `Transfer.is_idle()` and `Transfer.is_rx_not_empty()` for `serial::Rx` and `uart::Rx`.
* Relax trait bounds this methods.

See #586 for details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 [#617]: https://github.com/stm32-rs/stm32f4xx-hal/pull/617
 [#618]: https://github.com/stm32-rs/stm32f4xx-hal/pull/618
 [#623]: https://github.com/stm32-rs/stm32f4xx-hal/pull/623
+### Changed
+
+### Added
+
+ - Improve SPI::new* docs [#587]
+ - Implement `serial::RxISR` for `dma::Transfer<..., PERIPHERAL, ...>` where `PERIPHERAL: serial::RxISR`, add `rtic-serial-dma-rx-idle` example [#588]
+
+### Fixed
 
 ## [v0.15.0] - 2023-03-13
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -416,6 +416,10 @@ name = "rtic-i2s-audio-in-out"
 required-features = ["stm32f411", "i2s", "rtic"]
 
 [[example]]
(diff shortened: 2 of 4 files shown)


qwerty19106:
It is ready for review.
I will add `serial-dma-rx.rs` a little later.

qwerty19106:
I add `rtic-serial-dma-rx-idle` example

burrbull:
I'm not happy in existence of `fn peripheral` which was introduced in https://github.com/stm32-rs/stm32f4xx-hal/pull/396 as it potentially break API safety.
I still think this was a mistake.

cc @therealprof your comments?

qwerty19106:
The `fn peripheral` problem seems to be delayed.
I can remove `fn peripheral` part of this PR to expedite the acceptance of this PR.
@burrbull

rursprung:
FYI: i had introduced the API you've marked as deprecated in #556 (see #550 for background). the original ticket also describes why i did this with a `ReadBuffer` instead of a `WriteBuffer` (it was a read use-case for me). if there was something wrong with that then i'm sorry (i'm absolutely no expert on STM32, HALs or this specific repo). but it'd just be good if the use-case i described there will still work one way or the other with a clean API afterwards.

qwerty19106:
@rursprung 
A [WriteBuffer](https://docs.rs/embedded-dma/latest/embedded_dma/trait.WriteBuffer.html) is 
> Trait for buffers that can be given to DMA for writing.

Moreover the [WriteBuffer::write_buffer](https://docs.rs/embedded-dma/latest/embedded_dma/trait.WriteBuffer.html#tymethod.write_buffer) method takes mutable reference unlike `ReadBuffer`:
> unsafe fn write_buffer(&self) -> (*mut Self::Word, usize);

Thus DMA read `word` from peripheral and write to `WriteBuffer`.
We should use `WriteBuffer` for `PeripheralToMemory`, and `ReadBuffer` for `MemoryToPeripheral`.

Probably your example compiles with this bug because your `buffer` implements both `WriteBuffer` and `ReadBuffer`.

burrbull:
Maybe it is better to implement `RxISR` for `Transfer` instead of using inherent methods.

qwerty19106:
I removed the controversial `peripheral()` code and add `RxISR` implementation for `Transfer`.

burrbull (approved):

PR #598, opened 2023-04-15
BREAKING CHANGE: Ability to use 1 DMA channel for TX or RX only for I2C dma implementation

### LINKED ISSUE #590: Single channel DMA support for I2C
Now the i2c DMA can only be used for both tx and rx, could we just use it for a single channel (e.g. for tx or rx only, and the other channel can still be used as normal mode)?

SpeedCrash100:
# Short description
This PR closes #590 

* Adds two methods to the blocking I2c implementation: `use_dma_tx`, `use_dma_rx` to create I2CMasterDma that use only one of Tx/Rx channels.
* Clients can only use `write_dma` and all blocking methods if using Tx only I2c DMA. 
* Clients can only use `read_dma` and all blocking methods if using Rx only I2c DMA. 

# What has changed for clients
* An `I2CMasterDma` signature now has 3 generic parameters:
  1.  I2C instance
  2. `TxDMA` with same I2C instance, DMA stream and channel number or `NoDMA` if Tx is not used
  3.  `RxDMA` with same I2C instance, DMA stream and channel number or `NoDMA` if Rx is not used
* The client must import `I2CMasterHandleIT` from `stm32f4xx_hal::i2c::dma` for access interrupt handling methods
* The client must also import `TxDMA`, `RxDMA`, `NoDMA` from same module


burrbull:
useful traits could be reexported in `prelude`

SpeedCrash100:
> useful traits could be reexported in `prelude`

Done. I'm searching for way to dedublicate `write_dma` and `read_dma` methods for now and I have tested code on example only. So PR will be draft for now.

burrbull:
Please, never do "merge master" in pull request. Use `rebase` instead.

SpeedCrash100:
Ok. I have tested Tx and Tx/Rx modes. I don't know applications for Rx only mode so this part not tested.

burrbull:
cc @Leonana69 is this what you've asked for?

burrbull (approved):

Leonana69:
> cc @Leonana69 is this what you've asked for?

Yes, thanks so much!

PR #604, opened 2023-04-18
generic alt enums + pwm channel builders

burrbull:
cc @therealprof 

cc @janschiefer
I planning to change API for passing pins in PWM constructors. Now you need to specify channel number manually. See examples:
https://github.com/stm32-rs/stm32f4xx-hal/blob/alt-pwm-simple/examples/pwm-dead-time.rs
https://github.com/stm32-rs/stm32f4xx-hal/blob/alt-pwm-simple/examples/pwm.rs
Could you test, please?


therealprof (approved):
LGTM, thanks.

bors r+

Issue #624, opened 2023-05-03
SPI Master is fully broken now because mstr bit is cleared


qwerty19106:
@burrbull 
It changed in [#609](https://github.com/stm32-rs/stm32f4xx-hal/pull/609).

First `Spi::pre_init` set `mstr` bit:
```rust
impl<SPI: Instance, const BIDI: bool, W> Spi<SPI, BIDI, W> {
    /// Pre initializing the SPI bus.
    fn pre_init(self, mode: Mode, freq: Hertz, clock: Hertz) -> Self {
        // disable SS output
        self.spi.cr2.write(|w| w.ssoe().clear_bit());

        let br = match clock.raw() / freq.raw() {
            0 => unreachable!(),
            1..=2 => 0b000,
            3..=5 => 0b001,
            6..=11 => 0b010,
            12..=23 => 0b011,
            24..=47 => 0b100,
            48..=95 => 0b101,
            96..=191 => 0b110,
            _ => 0b111,
        };

        self.spi.cr1.write(|w| {
            w.cpha().bit(mode.phase == Phase::CaptureOnSecondTransition);
            w.cpol().bit(mode.polarity == Polarity::IdleHigh);
            // mstr: master configuration
            w.mstr().set_bit();
            w.br().bits(br);
            // lsbfirst: MSB first
            w.lsbfirst().clear_bit();
            // ssm: enable software slave management (NSS pin free for other uses)
            w.ssm().set_bit();
            // ssi: set nss high
            w.ssi().set_bit();
            w.rxonly().clear_bit();
            // dff: 8 bit frames
            w.dff().clear_bit()
        });

        self
    }
}
```
Next `Spi::init` clear `mstr` bit:
```rust
impl<SPI: Instance, const BIDI: bool, W: FrameSize> Spi<SPI, BIDI, W> {
    pub fn init(self) -> Self {
        self.spi.cr1.modify(|_, w| {
            // bidimode: 2-line or 1-line unidirectional
            w.bidimode().bit(BIDI);
            w.bidioe().bit(BIDI);
            // master/slave mode
            w.mstr().clear_bit();
            // data frame size
            w.dff().bit(W::DFF);
            // spe: enable the SPI bus
            w.spe().set_bit()
        });

        self
    }
}
```

Probably we should remove `w.mstr().clear_bit();` line from `Spi::init`.

burrbull:
[...]

PR #609, opened 2023-04-20
split master/slave spi structures

burrbull:

PR #625, opened 2023-05-03
Fix mstr bit for SPI Master/Slave

qwerty19106:
See #624 for details.


burrbull:
I don't think full deleting of those lines is good idea.
I'd just change `clear_bit => set_bit` and vise versa.

qwerty19106:
We call `clear_bit` and `set_bit` by `pre_init` already.
And before #609 we not call `clear_bit` and `set_bit` in `pre_init`.

burrbull (approved):

Issue #630, opened 2023-05-07
The `serial::Rx` and `uart::Rx` splitting causes copypaste for high orger API


qwerty19106:
@burrbull
The `serial` and `uart` was splitted in #608.

I read bytes by DMA until IDLE interrupt was received. I create support struct to reduce copypaste:
```Rust
pub struct RxDmaIdle<STREAM, const CHANNEL: u8, USART, const BUFFER_SIZE: usize>
where
    STREAM: Stream,
    USART: Instance,
    Rx<USART>: PeriAddress<MemSize = u8>,
{
    transfer: Transfer<STREAM, CHANNEL, Rx<USART>, PeripheralToMemory, RxBufferObject<BUFFER_SIZE>>,
    pool: &'static ObjectPool<Vec<u8, BUFFER_SIZE>>,
}

impl<STREAM, const CHANNEL: u8, USART, const BUFFER_SIZE: usize>
    RxDmaIdle<STREAM, CHANNEL, USART, BUFFER_SIZE>
where
    STREAM: Stream,
    ChannelX<CHANNEL>: Channel,
    USART: Instance,
    Rx<USART>: PeriAddress<MemSize = u8> + RxISR + DMASet<STREAM, CHANNEL, PeripheralToMemory>,
    ...
{
	pub fn new(
        mut rx: Rx<USART>,
        stream: STREAM,
		...)  -> Self {
			rx.unlisten();
        	rx.listen_idle();	

			...
		}

	pub fn uart_interrupt_impl(
        &mut self,
    ) -> Result<RxBufferObject<BUFFER_SIZE>, RxPoolExhaustedError> {
		...
	}
}
```

But after #608 I should duplicate it for `serial::Rx` and `uart::Rx` because the `unlisten()` and `listen_idle()` is methods of `Rx`, not trait.

It would be logical to have one struct `Rx` for `USART` and `UART`, some common traits with common/different implementation and different traits with different implementation.
The `Tx` have the same problem.

Are there really important reasons to split `serial::Rx` and `uart::Rx`?

<details>
<summary>Full RxDmaIdle code</summary>
```Rust
use embedded_dma::WriteBuffer;
use heapless::{pool::object_simple::*, Vec};
use stm32f4xx_hal::{
    dma::{
        config::DmaConfig,
        traits::{Channel, DMASet, PeriAddress, Stream},
        ChannelX, PeripheralToMemory, Transfer,
    },
[...]

PR #631, opened 2023-05-07
listen trait

burrbull:


burrbull:
cc @qwerty19106

qwerty19106:
I will try rewrite `seial` and `uart` modules to fix #630.
It require a few days for experiments.

burrbull:
Close in favor of #673

PR #637, opened 2023-05-11
Dual i2s core

YruamaLairba:
Hi, i'm trying to implement the 'device' that is suitable for full duplex i2s communication. I'd like to have opinion about implementation, i'm not sure it follow the general style of the crate. I don't have example to follow, I'm in a unusual situation where the 'device' is composed by 2 peripherals.

### DIFF
--- a/src/i2s.rs
+++ b/src/i2s.rs
@@ -6,7 +6,7 @@
 //! these chips because their `I2S2EXT` and `I2S3EXT` peripherals are missing from their package
 //! access crate.
 
-use crate::gpio::{self, NoPin};
+use crate::gpio::{self, NoPin, PinSpeed, Speed};
 use crate::pac;
 #[allow(unused)]
 use crate::rcc::{self, Clocks, Reset};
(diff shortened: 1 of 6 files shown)


burrbull:
looks goog for me

YruamaLairba:
I ask that because i'd like to have a similar style to the single i2s, but the current implementation feel weird, espacially if i want to implement the equivalent of I2sExt. Anyway, i can't change it later. At the moment i just need a way to test stuff in  stm32_i2s crate

src/i2s.rs
```diff
@@ -26,6 +26,12 @@ pub trait Instance:
 {
 }
 
+/// Trait for SPI peripheral that have an extension for full duplex i2s capability.
+pub trait Extension: gpio::alt::I2sExtPin {
+    /// The I2SEXT peripheral that extend the SPI peripheral
+    type I2sExtPeripheral;
+}
+
```
burrbull, line 34: Possibly it is better to depend `DualI2sPeripheral` on `I2sPeripheral` and same:

```suggestion
pub trait DualInstance: Instance + gpio::alt::I2sExtPin {
    type I2sExtPeripheral;
}

```

YruamaLairba:
I, can i have a comment about this work ? should I rebase ?

burrbull:
> I, can i have a comment about this work ? should I rebase ?

Yes, rebase, please.

burrbull:
After rebase you can remove speed workaround:
See  https://github.com/stm32-rs/stm32f4xx-hal/commit/d8905f2232e5e545c87d6aa8ad54114143272b0b

YruamaLairba:
rebased! 

@burrbull: sorry, i didn't understood  if you plan to apply your patch yourself ? in this case i don't need to do anything, but please comment why you are changing pin speed.

burrbull:
Too many commits. Squash, please

burrbull (approved):

PR #640, opened 2023-05-14
extended 64-bit monotonic timer

burrbull:


examples/rtic-tick.rs
```diff
@@ -20,8 +23,10 @@ mod app {
         led: PC13<Output>,
     }
 
-    #[monotonic(binds = TIM2, default = true)]
-    type MicrosecMono = MonoTimerUs<pac::TIM2>;
+    //#[monotonic(binds = TIM2, default = true)]
+    //type MicrosecMono = MonoTimerUs<pac::TIM2>;
```
therealprof, line 27: Seems duplicate?

examples/rtic-tick.rs
```diff
@@ -30,8 +35,10 @@ mod app {
 
         let gpioc = ctx.device.GPIOC.split();
         let led = gpioc.pc13.into_push_pull_output();
+        defmt::info!("Start");
 
-        let mono = ctx.device.TIM2.monotonic_us(&clocks);
+        //let mono = ctx.device.TIM2.monotonic_us(&clocks);
```
therealprof, line 40: Same here.

examples/rtic-tick.rs
burrbull: no, it is not

therealprof: I mean, what is the purpose of having a commented block of 2 lines that is identical to the 2 lines before?

burrbull: Those "2 lines before" are deleted lines. Red.

therealprof (approved):
bors r+

PR #645, opened 2023-05-27
Add a basic blocking QSPI interface

X-yl:
This module implements the QuadSPI interface which allows high speed communication with external flash memory.

Limitations:
    - Interrupts are not supported.
    - Status polling mode is not supported.

### DIFF
--- a/examples/qspi-w25q.rs
+++ b/examples/qspi-w25q.rs
@@ -1,0 +1,180 @@
+//! Example of using the QSPI peripheral with a W25Q flash chip (W25Q128JV).
+//! Pins configured for QSPI Bank1 of STM32F412 board. Adjust as needed.
+
+#![no_std]
+#![no_main]
+
+use cortex_m_rt::entry;
+use cortex_m_semihosting::hprintln;
+use panic_semihosting as _;
+use stm32f4xx_hal as hal;
+use stm32f4xx_hal::gpio::GpioExt;
+use stm32f4xx_hal::qspi::{
+    FlashSize, MemoryMapped, Qspi, QspiConfig, QspiError, QspiMemoryMappedConfig, QspiMode,
+    QspiPins, QspiReadCommand, QspiWriteCommand,
+};
+
+pub struct W25Q<PINS: QspiPins> {
+    qspi: Qspi<PINS>,
+}
+
+pub struct DeviceId(u8);
+
+impl<PINS> W25Q<PINS>
+where
+    PINS: QspiPins,
+{
+    pub fn new(qspi: Qspi<PINS>) -> Result<Self, QspiError> {
+        let mut chip = Self { qspi };
+        chip.release_from_power_down()?;
+        chip.quad_enable()?;
+        Ok(chip)
+    }
+
+    pub fn release_from_power_down(&mut self) -> Result<DeviceId, QspiError> {
+        let mut buf = [0u8; 1];
+
+        self.qspi.indirect_read(
+            QspiReadCommand::new(&mut buf, QspiMode::SingleChannel)
+                .instruction(0xAB, QspiMode::SingleChannel)
+                .address(0x0, QspiMode::SingleChannel),
+        )?;
+
+        Ok(DeviceId(buf[0]))
+    }
+
+    pub fn wait_on_busy(&mut self) -> Result<(), QspiError> {
+        let mut buf = [0u8; 1];
+
+        loop {
+            self.qspi.indirect_read(
+                QspiReadCommand::new(&mut buf, QspiMode::SingleChannel)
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -154,6 +154,8 @@ pub mod prelude;
 #[cfg(feature = "device-selected")]
 pub mod qei;
 #[cfg(feature = "device-selected")]
+pub mod qspi;
```
burrbull, line 157:
```suggestion
#[cfg(feature = "quadspi")]
pub mod qspi;
```

src/qspi.rs
```diff
@@ -0,0 +1,783 @@
...
+                .bit(command.double_data_rate)
+                .dcyc()
+                .bits(command.dummy_cycles)
+                .dmode()
+                .bits(command.data_mode as u8)
+                .imode()
+                .bits(imode)
+                .instruction()
+                .bits(instruction)
+        });
```
burrbull, line 586: You can do this more compact
```suggestion
        self.qspi.ccr.modify(|_, w| unsafe {
            w.fmode().bits(0b11 /* Memory mapped mode */);
            w.admode().bits(command.address_mode as u8);
            w.adsize().bits(self.config.address_size as u8);
            w.abmode().bits(abmode);
            w.absize().bits(absize);
            w.ddrm().bit(command.double_data_rate);
            w.dcyc().bits(command.dummy_cycles);
            w.dmode().bits(command.data_mode as u8);
            w.imode().bits(imode);
            w.instruction().bits(instruction)
        });```

File: src/qspi.rs [hunk 3]
```diff
@@ -0,0 +1,783 @@
+//! QSPI interface
+//!
+//! This module implements the QuadSPI interface which allows high speed
+//! communication with external flash memory.
+//!
+//! Limitations:
+//!     - Interrupts are not supported.
```
burrbull, line 7:
```suggestion
//!     - Quad mode only.
//!     - Interrupts are not supported.
```

src/qspi.rs
```diff
@@ -0,0 +1,783 @@
...
+        alt::quadspi::Bk2Io3,
+        alt::quadspi::Clk,
+    );
+}
+
+pub struct Qspi<Pins: QspiPins> {
+    qspi: QUADSPI,
+    config: QspiConfig,
+    _marker: PhantomData<Pins>,
+}
```
burrbull, line 53: I'd rename `Pins` -> `BANK` and keep pins inside. This allows you to add `release` for `Qspi`.
```suggestion
pub struct Qspi<BANK: QspiPins> {
    qspi: QUADSPI,
    config: QspiConfig,
    pins: BANK::Pins,
}
```

X-yl:
Thanks for your suggestions, I've amended the commit to incorporate them. Let me know if any other changes are required :)

X-yl:
Added `#[cfg(feature = "quadspi")]` to the RccBus stuff.

burrbull:
Could you add example?

burrbull:
> Thanks for your suggestions, I've amended the commit to incorporate them. Let me know if any other changes are required :)

https://github.com/stm32-rs/stm32f4xx-hal/commit/6650964c084cb8f3cb5d6ad8e006f22bcfb4b44b

X-yl:
Applied your change and added an example using a W25Q series flash chip.

X-yl:
(Adjusted clock prescaler for max speed)

burrbull:
Is `Speed::VeryHigh` required by QSPI / default?

X-yl:
> Is `Speed::VeryHigh` required by QSPI / default?

I don't think it's required, but that was the configuration generated by CubeMX so I used it as is. From my testing, VeryHigh is required for prescaler zero but it works without setting the speed for higher prescaler.

burrbull:
> I don't think it's required, but that was the configuration generated by CubeMX so I used it as is. From my testing, VeryHigh is required for prescaler zero but it works without setting the speed for higher prescaler.

https://github.com/stm32-rs/stm32f4xx-hal/commit/516a0091a21f96bdb70825398754d364aeb42563

X-yl:
Fixed clippy suggestions

examples/qspi-w25q.rs
```diff
@@ -0,0 +1,180 @@
...
+
+pub struct W25Q<PINS: QspiPins> {
+    qspi: Qspi<PINS>,
+}
+
+pub struct DeviceId(u8);
+
+impl<PINS> W25Q<PINS>
+where
+    PINS: QspiPins + 'static,
```
burrbull, line 25:
```suggestion
    PINS: QspiPins,
```

burrbull (approved):

PR #650, opened 2023-05-30
Stream x no generic

YruamaLairba:
Just to illustrate a discussion about StreamX

### DIFF
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -50,6 +50,25 @@ impl<T> Debug for DMAError<T> {
     }
 }
 
+/// Possible Channel of a DMA Stream.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DmaChannel {
+    Channel0 = 0,
+    Channel1 = 1,
+    Channel2 = 2,
+    Channel3 = 3,
+    Channel4 = 4,
+    Channel5 = 5,
+    Channel6 = 6,
+    Channel7 = 7,
+}
+
+impl Bits<u8> for DmaChannel {
+    fn bits(self) -> u8 {
+        self as u8
+    }
+}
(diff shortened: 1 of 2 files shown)


YruamaLairba:
what ? i have commit duplication ???

burrbull:
I still don't see explanation why you need this except now you can pass incorrect values in safe API.

YruamaLairba:
> I still don't see explanation why you need this except now you can pass incorrect values in safe API.

I don't need this, i just don't understand why you need generic here. Can you tell me where in the API and how  it enable incorrect value ?

burrbull:
> I don't need this, i just don't understand why you need generic here. Can you tell me where in the API and how it enable incorrect value ?

Your code don't do anything wrong.
But as `Stream` trait is public (should it be?), anyone can call `set_channel` with any `DmaChannel`. It is not my API, so I know not much more than you.

YruamaLairba:
> anyone can call `set_channel` with any `DmaChannel`

But this doesn't enable incorrect value . With `fn set_channel<const C: u8>(&mut self)`, it's already possible to give any Channel number  between 0 and 7 without restriction. I Also read again the datasheet, i realized there is'nt really invalid channel values, There is just stupid values depending the situation

burrbull:
> With `fn set_channel<const C: u8>(&mut self)`, it's already possible to give any Channel number between 0 and 7 without restriction.

You are right. This is hole in design.

burrbull:
F413 has channels 8 and 9 on DMA2

YruamaLairba:
@burrbull i just checked datasheet, there is actually 16 Channel for stm32f413. The higher channels just don't have peripheral connected to them

burrbull:
изображение

8 and 9 are used

PR #653, opened 2023-06-07
update embedded_hal 1 to alpha.10

burrbull:
TODO: add missing functions for SPI and PWM
FMPI2c::transaction


Cargo.toml
rursprung, line 1: due to semver and the way cargo handles semver-based dependencies these updates are not necessary (the `Cargo.lock` will automatically contain the newer versions if you build it fresh or run `cargo update`). forcing the update here is only necessary if you rely on a feature (or bugfix) introduced in the newer version. is this the case here?

PR #655, opened 2023-06-21
fix #604

### LINKED ISSUE #654: Example pwm.rs from v0.16.0 does not work on F401
Hi everybody!
Thanks for this library. I've just started learning rust and wanted to start from the most recent versions of libraries.
I tried to run pwm.rs example and it compiles (without any warning), but there is no activity on PA8 (or PA9) (checked by oscilloscope). After couple tries, I switched to ^0.15.0 and it worked just like that.
So there is some error in example for 0.16 or some issue with the library.

burrbull:
Fixes #654 

cc @KarolNi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+ - Fix #604 pwm output [#655]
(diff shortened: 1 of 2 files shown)


therealprof (approved):
LGTM

burrbull:
something with rustfmt ci

PR #666, opened 2023-07-03
Stream x improve

YruamaLairba:
Hi, i'm trying to rewrite the stream struct API, here is the list modification i'm doing on stream:
- implement missing feature (already discussed in another Draft Pull Request)
- remove unecessary generics (lready discussed elsewhere)
- replace "class" method by "object" method to follow "peripheral ownership pattern"
- remove "get" prefix on getter to follow coding style of rust compiler
- rename things, to give a more natural, shorter or less ambiguous name
- replace some raw parameter by enum to make thing safer or more natural
- disable method just write 0 to the enable bit. In my opinion, trying to handle thing when disabling  the hardware is the 
- role of an higher level abstraction

This is what i'm doing on Transfer struct
-adapt to make it compile with the modified stream API
-put the "disable" logic i removed from Stream
-EDIT add or fix method for interrupt flags manipulations, this is required to make i2c/dma module to compile again

EDIT: i2c/dma module use transfer and stream, so fixing it to fit the change

### DIFF
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -50,15 +50,113 @@ impl<T> Debug for DMAError<T> {
     }
 }
 
+// most of STM32F4 have 8 DmaChannel
+#[cfg(not(feature = "gpio-f413"))]
+/// Possible Channel of a DMA Stream.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DmaChannel {
+    Channel0 = 0,
+    Channel1 = 1,
+    Channel2 = 2,
+    Channel3 = 3,
+    Channel4 = 4,
+    Channel5 = 5,
+    Channel6 = 6,
+    Channel7 = 7,
+}
+
+//  STM32F413 and STM32F423 have 16 DmaChannel
+#[cfg(feature = "gpio-f413")]
+/// Possible Channel of a DMA Stream.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DmaChannel {
+    Channel0 = 0,
+    Channel1 = 1,
+    Channel2 = 2,
+    Channel3 = 3,
+    Channel4 = 4,
+    Channel5 = 5,
+    Channel6 = 6,
+    Channel7 = 7,
+    Channel8 = 8,
+    Channel9 = 9,
+    Channel10 = 10,
+    Channel11 = 11,
+    Channel12 = 12,
+    Channel13 = 13,
+    Channel14 = 14,
+    Channel15 = 15,
+}
+
+impl Bits<u8> for DmaChannel {
+    fn bits(self) -> u8 {
+        self as u8
+    }
+}
+
+/// Peripheral increment offset size (pincos)
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PeripheralIncrementOffset {
+    /// The offset size is linked to the peripheral data size (psize)
+    PeripheralDataSize = 0,
+    /// The offset size is fixed to 4 (32 bits alignement)
+    FixedSize = 1,
+}
+
+impl Bits<bool> for PeripheralIncrementOffset {
+    fn bits(self) -> bool {
+        match self {
+            PeripheralIncrementOffset::PeripheralDataSize => false,
+            PeripheralIncrementOffset::FixedSize => true,
+        }
+    }
+}
+
+/// Size of data transfered during a dma stream request
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DmaDataSize {
+    Byte = 0,
+    HalfWord = 1,
+    Word = 2,
+}
+
+impl Bits<u8> for DmaDataSize {
+    fn bits(self) -> u8 {
+        self as u8
+    }
+}
+
 /// Possible DMA's directions.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum DmaDirection {
     /// Memory to Memory transfer.
-    MemoryToMemory,
+    MemoryToMemory = 2,
     /// Peripheral to Memory transfer.
-    PeripheralToMemory,
+    PeripheralToMemory = 0,
     /// Memory to Peripheral transfer.
-    MemoryToPeripheral,
+    MemoryToPeripheral = 1,
+}
+
+impl Bits<u8> for DmaDirection {
+    fn bits(self) -> u8 {
+        self as u8
+    }
+}
+
+/// Dma flow controller selection
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DmaFlowController {
+    Dma = 0,
+    Peripheral = 1,
+}
+
+impl Bits<bool> for DmaFlowController {
+    fn bits(self) -> bool {
+        match self {
+            DmaFlowController::Dma => false,
+            DmaFlowController::Peripheral => true,
+        }
+    }
 }
 
 /// DMA from a peripheral to a memory location.
@@ -185,7 +283,7 @@ pub enum CurrentBuffer {
     /// The first buffer (m0ar) is in use.
     FirstBuffer,
     /// The second buffer (m1ar) is in use.
-    DoubleBuffer,
+    SecondBuffer,
 }
 
 impl Not for CurrentBuffer {
@@ -193,9 +291,53 @@ impl Not for CurrentBuffer {
 
     fn not(self) -> Self::Output {
         if self == CurrentBuffer::FirstBuffer {
-            CurrentBuffer::DoubleBuffer
+            CurrentBuffer::SecondBuffer
         } else {
             CurrentBuffer::FirstBuffer
+        }
+    }
+}
+
+/// Structure representing setup of common interrupts.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct DmaCommonInterrupts {
+    pub transfer_complete: bool,
+    pub half_transfer: bool,
+    pub transfer_error: bool,
+    pub direct_mode_error: bool,
+}
+
+impl DmaCommonInterrupts {
+    /// Return a new DmaCommonInterrupts with all fields set to true.
+    const fn all() -> Self {
+        Self {
+            transfer_complete: true,
+            half_transfer: true,
+            transfer_error: true,
+            direct_mode_error: true,
+        }
+    }
+}
+
+/// Structure returned by Stream or Transfer all_flags() method.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct DmaFlags {
+    pub transfer_complete: bool,
+    pub half_transfer: bool,
+    pub transfer_error: bool,
+    pub direct_mode_error: bool,
+    pub fifo_error: bool,
+}
+
+impl DmaFlags {
+    /// Return a new DmaFlags with all fields set to true.
+    const fn all() -> Self {
+        Self {
+            transfer_complete: true,
+            half_transfer: true,
+            transfer_error: true,
+            direct_mode_error: true,
+            fifo_error: true,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/dma/mod.rs
```diff
@@ -608,79 +608,95 @@ macro_rules! dma_stream {
...
+                fn clear_fifo_error_flag(&mut self) {
                     //NOTE(unsafe) Atomic write with no side-effects and we only access the bits
                     // that belongs to the StreamX
                     let dma = unsafe { &*I::ptr() };
                     dma.$ifcr.write(|w| w.$feif().set_bit());
                 }
 
                 #[inline(always)]
-                fn get_transfer_complete_flag() -> bool {
+                fn all_flags(&self) -> (bool, bool, bool, bool, bool)
```
burrbull, line 689: Maybe better return some structure. Tuple of booleans say nothing about what these flags mean.

burrbull: or use `bitflags`

YruamaLairba: Agree, i will use a structure to keep things simple. Will do the same for other function that get or set several bits at once.

Have you an option about 'interupts' and 'set_interrupt' function ?

Anyway, my priority is to adapt code to make it compile. Stream and transfer are used in i2c module and the code here is also weird. I think this require to implement many Stream method on Transfer struct.

I also noticed that Stream and StreamISR trait may unnecessary. I mean, this is only implemented for <`StreamX<DMA, S> and implementation are already constrained to valid stream by `Sealed` ( but it 's better to use a `StreamInstance` marker trait to make the intend clear). However, I wil not change that.

YruamaLairba: Yeees, the lib compile again

YruamaLairba:
can't rebase, there is conflicting change

burrbull:
first squash commits and then rebase
See https://github.com/stm32-rs/stm32f4xx-hal/commits/StreamX_improve

YruamaLairba:
will do, but ii need to fix examples first.

YruamaLairba:
@burrbull do you know a way to check all examples ? when i use `cargo check --examples --feature="stm32f411"` only examples that don't require additional feature are checked. For the moment i just manually checked examples that contain "dma" in the name.

burrbull:
if you on linux you could run `./tools/check.py` to check all chips. Although it is slightly outdated (for example doesn't have `rtic` feature).

YruamaLairba:
merging correctly will be difficult, there was many change since i started this PR, i fear Git will automatically resolve conflict in incoherent way. I have a example not working but I may forgot to check it earlier

YruamaLairba:
@burrbull ready for a review, i just didn't change the CHANGELOG, do know what description i should put. I'm very skeptical Transfer. The main problem is the ownership taking of the peripheral, this prevent to freely use the peripheral. But beyond that, it's seems some people already added some functionality to it just to make it usable for some peripheral, which is again the intend of being general purpose stuff.

YruamaLairba:
Sorry, I just found a typo i repeated many time because of automatic completion

src/dma/traits.rs
```diff
@@ -19,64 +19,76 @@ pub trait SafePeripheralRead {}
...
+    /// The tuple contain in order:
+    ///  - transfer complete flag
+    ///  - half transfer flag
+    ///  - transfer error flag
+    ///  - direct mode error flag
+    ///  - fifo_error flag
+    fn all_flags(&self) -> DmaFlags;
 
     /// Get transfer complete flag.
-    fn get_transfer_complete_flag() -> bool;
```
burrbull, line 41: I understand the reason of deleting prefix, but it is hard to read code without it.
Possibly it is better to use `is_` prefix here for consistency with PAC's bit reading names.

src/dma/mod.rs
```diff
@@ -185,21 +283,38 @@ pub enum CurrentBuffer {
...
+}
+
+/// Structure returned by Stream or Transfer all_flags() method.
+pub struct DmaFlags {
+    pub transfer_complete: bool,
+    pub half_transfer: bool,
+    pub transfer_error: bool,
+    pub direct_mode_error: bool,
+    pub fifo_error: bool,
+}
```
burrbull, line 316:
```suggestion
pub struct DmaFlags {
    pub TransferComplete: bool,
    pub HalfTransfer: bool,
    pub TransferError: bool,
    pub DirectModeError: bool,
    pub FifoError: bool,
}
```
Same for `DmaCommonInterrupts`

burrbull: Even better to use bitflags here:
https://github.com/YruamaLairba/stm32f4xx-hal/pull/1

src/dma/traits.rs
```diff
@@ -86,80 +98,82 @@ pub trait Stream: StreamISR + crate::Sealed {
...
-    fn get_interrupts_enable() -> (bool, bool, bool, bool);
+    fn set_direction(&mut self, direction: DmaDirection);
+
+    /// Set the flow controller (pfctrl).
+    fn set_flow_controller(&mut self, value: DmaFlowController);
+
+    /// Convenience method to configure several interrupts of the DMA stream.
+    ///
+    /// Note: fifo_error interrupt is not concerend because it's in a different register
+    fn set_common_interrupts(&mut self, interrupts: DmaCommonInterrupts);
```
burrbull, line 155: for other peripherals we use `listen/unlisten` here.

src/dma/mod.rs
YruamaLairba: Why ? Doing that will trigger a warning!

burrbull: oh. sorry. this is field

src/dma/traits.rs
YruamaLairba: maybe i can also remove the "_flag" suffix ?
```Rust
fn is_transfer_complete(&self) -> bool;
fn clear_transfer_complete(&mut self) ->bool;
```

burrbull: Yes. Summarizing:
`is_smth`
`clear_smth`
`listen_smth`
`unlisten_smth`

burrbull: cc @therealprof

src/dma/traits.rs
YruamaLairba: for this one, I suggest this
```Rust
/// Listen/Unlisten several interrupts of the DMA stream.
fn listen_common_interrupts(&mut self, interrupts: DmaCommonInterrupts);
```

for individual interrupts setting i can use listen/unlisten methods

burrbull: If you are using `DmaCommonInterrupts` as structure, yes.

I would use `bitflags` instead of simple structure here.
Advantage is you can do `OR` (`|`) operations on flags.

Difference is in semantics.
When you `clear` the flag other flags are unchanged. Same for `bitflags` + both independent `listen` and `unlisten`.

Current `set_interrupts_enable`/`set_common_interrupts` behavior is different. You need to set/reset all flags, not just that you expect to be changed.

Although I can be wrong. This is breaking change.

And I'm not sure which variant can be better optimized.

src/dma/traits.rs
YruamaLairba: done for all interrupt individually

src/i2c/dma.rs
```diff
@@ -597,22 +597,22 @@ where
...
 
                 self.finish_transfer_with_result(Err(Error::TransferError));
 
                 return;
             }
 
-            if TX_STREAM::get_transfer_complete_flag() {
-                tx_t.clear_transfer_complete_interrupt();
+            if tx_t.is_transfer_complete() {
+                tx_t.clear_transfer_complete();
```
burrbull, line 615: could be optimized with `all_flags`

therealprof (commented):
I can't really comment on the soundness of the DMA stuff but the approach looks good to me. 👍🏻

src/dma/traits.rs
YruamaLairba: done, i turned all  "set_smth_interrupt" in to set of "listen_smth and unlisten_smth"

YruamaLairba:
Since i Just started to use bitflag, i wonder if i should use it for all structure containing a set of boolean to keep coherence

YruamaLairba:
crap, i missed the annoying "[from_bits_retain](https://docs.rs/bitflags/2.2.1/bitflags/trait.BitFlags.html#tymethod.from_bits_retain)" method in bitflags. This make my code unsound :(

src/i2c/dma.rs
YruamaLairba: done

YruamaLairba:
@burrbull is it ok ? i think there may a lack of coherency with flags, but i found bitflag! a bit deceiving. The "from_bit_retain" method have disastrous consequence, and using the "Flag::A|Flag::B|Flag::C" is actually very verbose with concrete case:
"DmaCommonInterrupt::TRANSFER_COMPLETE|DmaCommonInterrupt::HALF_TRANSFER ..."

burrbull:
ok lets leave bitflags for another PR.

src/dma/mod.rs
```diff
@@ -50,15 +50,113 @@ impl<T> Debug for DMAError<T> {
...
+    Channel6 = 6,
+    Channel7 = 7,
+    Channel8 = 8,
+    Channel9 = 9,
+    Channel10 = 10,
+    Channel11 = 11,
+    Channel12 = 12,
+    Channel13 = 13,
+    Channel14 = 14,
+    Channel15 = 15,
```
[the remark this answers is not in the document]
burrbull: why you closed this? Channels 10-15 are useless.

YruamaLairba:
@burrbull i already commited thing using bitflags

src/dma/mod.rs
```diff
@@ -1361,34 +1592,50 @@ where
...
-        stream.set_transfer_complete_interrupt_enable(config.transfer_complete_interrupt);
-        stream.set_half_transfer_interrupt_enable(config.half_transfer_interrupt);
-        stream.set_transfer_error_interrupt_enable(config.transfer_error_interrupt);
-        stream.set_direct_mode_error_interrupt_enable(config.direct_mode_error_interrupt);
-        stream.set_fifo_error_interrupt_enable(config.fifo_error_interrupt);
+
+        // TODO fix that when are more conveninent way to set interrupt is available
+        use DmaCommonInterruptsMask as Mask;
+        let mut interrupts_mask = Mask::empty();
+        interrupts_mask.set(Mask::TRANSFER_COMPLETE, config.transfer_complete_interrupt);
```
burrbull, line 1628: as initialized with `empty` it is enough to use `insert` instead of `set`

burrbull:
@YruamaLairba could you look at/test #673 ?

YruamaLairba:
@burrbull what is missing to merge this PR ? do wait for having a good system to manipulate a set of flags ?

burrbull:
> @burrbull what is missing to merge this PR ? do wait for having a good system to manipulate a set of flags ?

I'm fine to leave bit flags for some future PR.

But I don't see sense in duplicates (like `DmaCommonInterrupts` and `DmaCommonInterrupts`). It's confusing.
Or both `DmaFlags` and `DmaCommonInterrupts` are structs for read and write or both are bit flags.
Remove `DmaCommonInterruptsMask`.

YruamaLairba:
So  finally, i should just ignore what you did in the Flags branch and use structure with boolean field to manipulate several interrupts or flags at once ?

burrbull:
> So finally, i should just ignore what you did in the Flags branch and use structure with boolean field to manipulate several interrupts or flags at once ?

Yes

YruamaLairba:
done; i also changed clear_all_flags(&mut self) by clear_flags(&mut self, flags:DmaFlags) and added some all() method to structure

burrbull:
Ok. I'll merge this after fix release https://github.com/stm32-rs/stm32f4xx-hal/pull/674

burrbull:
Add to `CHANGELOG.md`, please

YruamaLairba:
Is it ok or a squash is required ?

burrbull:
better to squash

YruamaLairba:
squashed

burrbull (approved):

PR #667, opened 2023-07-04
merge f7xx-hal

burrbull:


therealprof:
I find it weird to have F4 and F7 mixed into one since they're completely different series but if that's what people want, who am I to stop them. 🤪

burrbull:
> they're completely different series

F2/F4/F7 are very similar series. Peripherals on higher F4 chips (F446/69) are even closer to F7 then other F4.

therealprof:
> > they're completely different series
> 
> F2/F4/F7 are very similar series. Peripherals on higher F4 chips (F446/69) are even closer to F7 then other F4.

Yeah, there're some newer outliers. From that PoV all of the STM32 chips are very similar (sans the clock trees) because there're only a few versions of peripherals all the chips share...

eldruin:
I am fine with anything you would like to maintain. The F7 repository has seen very little action lately.

PR #673, opened 2023-07-18
Flags

burrbull:
Based on #666

### DIFF
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -16,14 +16,15 @@ use core::{
     sync::atomic::{compiler_fence, Ordering},
 };
 use embedded_dma::{ReadBuffer, WriteBuffer};
+use enumflags2::BitFlags;
 
 use crate::{pac, rcc};
 
 pub mod traits;
 use crate::serial::RxISR;
 use traits::{
-    sealed::Bits, Channel, DMASet, Direction, Instance, PeriAddress, SafePeripheralRead, Stream,
-    StreamISR,
+    sealed::Bits, Channel, DMASet, Direction, DmaEventExt, DmaFlagExt, Instance, PeriAddress,
+    SafePeripheralRead, Stream, StreamISR,
 };
 
 /// Errors.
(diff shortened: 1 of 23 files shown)


src/dma/mod.rs
```diff
@@ -524,63 +585,49 @@ where
...
-        DmaCommonInterrupts {
-            transfer_complete: cr.tcie().bit_is_set(),
-            half_transfer: cr.htie().bit_is_set(),
-            transfer_error: cr.teie().bit_is_set(),
-            direct_mode_error: cr.dmeie().bit_is_set(),
+    fn listen(&mut self, interrupts: DmaCommonInterrupts) {
+        unsafe {
+            Self::st()
+                .cr
+                .modify(|r, w| w.bits(r.bits() | interrupts.bits() as u32));
```
YruamaLairba, line 592: because of the existence of "from_bits_retain" value of unused bit is not guaranteed, that mean you need to "delete" those bit before the bitwise "or" operation

src/dma/mod.rs
```diff
@@ -524,63 +585,49 @@ where
...
 
     #[inline(always)]
-    fn set_transfer_complete_interrupt(&mut self, transfer_complete_interrupt: bool) {
-        unsafe { Self::st() }
-            .cr
-            .modify(|_, w| w.tcie().bit(transfer_complete_interrupt));
+    fn listen_only(&mut self, interrupts: DmaCommonInterrupts) {
+        unsafe {
+            Self::st().cr.modify(|r, w| {
+                w.bits(r.bits() & !DmaCommonInterrupts::MASK | interrupts.bits() as u32)
```
YruamaLairba, line 600: same remark as for "listen"

src/dma/mod.rs
```diff
@@ -524,63 +585,49 @@ where
...
     #[inline(always)]
-    fn set_half_transfer_interrupt(&mut self, half_transfer_interrupt: bool) {
-        unsafe { Self::st() }
-            .cr
-            .modify(|_, w| w.htie().bit(half_transfer_interrupt));
+    fn unlisten(&mut self, interrupts: DmaCommonInterrupts) {
+        unsafe {
+            Self::st()
+                .cr
+                .modify(|r, w| w.bits(r.bits() & !(interrupts.bits() as u32)));
```
YruamaLairba, line 610: same remark as for listen

src/dma/mod.rs
YruamaLairba, line 600: same remark as for listen

src/dma/mod.rs
```diff
@@ -636,131 +683,40 @@ where
...
-                    dma.$ifcr.write(|w| w.$dmeif().set_bit());
-                }
-
-                #[inline(always)]
-                fn clear_fifo_error_flag(&mut self) {
-                    //NOTE(unsafe) Atomic write with no side-effects and we only access the bits
-                    // that belongs to the StreamX
-                    let dma = unsafe { &*I::ptr() };
-                    dma.$ifcr.write(|w| w.$feif().set_bit());
+                    dma.$ifcr.write(|w| unsafe { w.bits(flags.bits() as u32) });
```
YruamaLairba, line 693: same remark as listen you need to to delete bit that doesn't correspond to a flag. And I think you missed to apply $isr_shift

src/dma/mod.rs
burrbull: you want to say that `from_bits_retain` must be `unsafe`?

src/dma/mod.rs
```diff
@@ -636,131 +683,40 @@ where
...
-                    let dma = unsafe { &*I::ptr() };
-                    dma.$isr.read().$dmeisr().bit_is_set()
-                }
-
-                #[inline(always)]
-                fn fifo_error_flag(&self) -> bool {
-                    //NOTE(unsafe) Atomic read with no side effects
-                    let dma = unsafe { &*I::ptr() };
-                    dma.$isr.read().$feisr().bit_is_set()
+                    DmaFlags::from_bits_retain(
```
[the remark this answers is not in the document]
burrbull: `from_bits_truncate` is not simple
https://docs.rs/bitflags/latest/src/bitflags/traits.rs.html#80-96

YruamaLairba:
@burrbull if i summarize, you removed single "listen_smth" in favor of "listen(DmaInterupts)" and now to check a flag you need to do stream.all_flags().is_smth() instead of stream.is_smth()

src/dma/mod.rs
burrbull: > you need to "delete" those bit before the bitwise "or" operation

tried to fix in last commit

YruamaLairba:
note for bitflags, i wonder it it worse to check for better alternative. for example with bitflag 1.3.* you have garantee that unused bit are delete.
I also saw bitbag wich define flags as enum, and a bitbag where you store your flags. It allow to do import flag name in the current scope and write thing like 
```rust
let flags = A | B | C;
```
however Bitbag may miss some functionnality compared to bitflags

burrbull:
> @burrbull if i summarize, you removed single "listen_smth" in favor of "listen(DmaInterupts)" and now to check a flag you need to do stream.all_flags().is_smth() instead of stream.is_smth()

No. It should still work.

src/dma/mod.rs
burrbull: missed this in previous commit

burrbull:
> > I also saw bitbag wich define flags as enum, and a bitbag where you store your flags. It allow to do import flag name in the current scope and write thing like
> 
> Interesting. Have not see it before.

Looks like it is not marked as `no_std`.

burrbull:
Looks like I've found good replacement.
https://docs.rs/flagset/0.4.3/flagset/

therealprof (approved):
Sweet, I like the new interface. Quite breaking, though.

Issue #696, opened 2023-11-22
UART Overrun error (interrupts) - how to increase buffer size?


hacknus:
I'm using an STM32F405 running the [freeRTOS wrapper](https://github.com/lobaro/FreeRTOS-rust) and I'm struggling to get UART working reliably.

I'm using:

```toml
[dependencies.stm32f4xx-hal] 
version = "0.18" 
 features = ["rt", "stm32f405", "usb_fs", "spi3"]
```

I'm using an interrupt to push the data into a ring buffer, which I then read out (this works!) but from time to time, I get the printouts
```shell
error uart read: Overrun
```
which to me seems like my interrupt misses some bytes that are received?

The messages I am receiving are 55 bytes long every 5 seconds.

```rust

let clocks = rcc
    .cfgr
    .use_hse(8.MHz())
    .sysclk(96.MHz())
    .hclk(48.MHz())
    .require_pll48clk()
    //.pclk1(24.MHz()) // APB1
    //.pclk2(24.MHz()) // APB2
    .freeze();

// initialize UART3 for RS485
let rx_3 = gpiod.pd9.into_alternate();
let tx_3 = gpiod.pd8.into_alternate();
let uart3 = Serial::new(
    dp.USART3,
    (tx_3, rx_3),
    Config::default()
        .baudrate(9600.bps())
        .parity_none()
        .stopbits(StopBits::STOP1),
    &clocks,
)
.unwrap();

uart3.listen(Event::RxNotEmpty); 
cortex_m::interrupt::free(|cs| {    
 *G_UART3.borrow(cs).borrow_mut() = Some(uart3);    
 *G_UART3_REDE.borrow(cs).borrow_mut() = Some(rede);    
 *G_UART3_BUFFER.borrow(cs).borrow_mut() = Some(Buffer::new());
 });


unsafe {
      usb_init(usb);
      cortex_m::peripheral::NVIC::unmask(Interrupt::OTG_FS);
      cortex_m::peripheral::NVIC::unmask(Interrupt::USART1);
      cortex_m::peripheral::NVIC::unmask(Interrupt::USART2);
      cortex_m::peripheral::NVIC::unmask(Interrupt::USART3);
      cortex_m::peripheral::NVIC::unmask(sw.interrupt());
  }
```

The interrupt handler looks like this:
```rust
#[interrupt]
#[allow(non_snake_case)]
fn USART3() {
    // put read value in ring buffer
[...]

PR #698, opened 2023-11-23
create uart-dma.rs example

hacknus:
I created a simple example for using UART with DMA (only on the rx side).
This solves the issue of overrun errors (see #696) when the interrupt handler is not quick enough and data is lost on the input.

### DIFF
--- a/examples/uart-dma.rs
+++ b/examples/uart-dma.rs
@@ -8,16 +8,20 @@ use stm32f4xx_hal::dma::{DmaFlag, PeripheralToMemory, Stream1};
 use core::cell::RefCell;
 use cortex_m::interrupt::Mutex;
 use cortex_m_rt::entry;
-use stm32f4xx_hal::pac::{interrupt, DMA1};
-use stm32f4xx_hal::pac::Interrupt;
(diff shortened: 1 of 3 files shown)


examples/uart-dma.rs
```diff
@@ -0,0 +1,245 @@
...
+fn DMA1_STREAM1() {
+    cortex_m::interrupt::free(|cs| {
+        if let Some(transfer) = G_TRANSFER.borrow(cs).borrow_mut().as_mut() {
+            // Its important to clear fifo errors as the transfer is paused until it is cleared
+            if transfer.flags() == DmaFlag::FifoError {
+                transfer.clear_fifo_error();
+            }
+            if transfer.flags() == DmaFlag::TransferComplete {
+                transfer.clear_transfer_complete();
+            }
```
burrbull, line 242: It is faster to clear without check, is not it?
```suggestion
            transfer.clear_flags(DmaFlag::FifoError | DmaFlag::TransferComplete);
```

hacknus: Agreed, so I guess it does not matter if we clear it even though the transfer is not yet fully complete (`HalfTransfer`) ?

burrbull (approved):

PR #700, opened 2023-11-24
Fix alternate function for LTDC pins PB0 PB1

ierturk:
There is conflict between the Table 10 and Table 12 in the STM32F429 datasheet.

PB0 and PB0 have alternate function 14 according to Table 10, but they are not documented on Table 12 correctly.

PB0 ->  LCD_R3
PB1 ->  LCD_R6

https://www.st.com/resource/en/datasheet/stm32f429zi.pdf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added an example to show how to do DMA with UART (Rx only) [#698]
 
+### Fixed
+- Missing alternate function definition for LTDC pins PB0 PB1
+
 ## [v0.18.0] - 2023-11-19
 
 ### Changed


burrbull:
Could you link (or screenshot) datasheet with these pinouts?

ierturk:
Here is the screenshots
table10
table12

burrbull:
are there other mistakes?

ierturk:
The other LTDC pins are correct.

burrbull (approved):

burrbull:
What about PB0/PB1 AF9?

burrbull:
Looks like I need to revert this PR as incorrect.

ierturk:
Do we need to correct the lines in LTDC with <9> as well? Without the fix, following error arises

into-alternate

burrbull:
I don't understand why you try use AF14 there. All datasheets say there should be AF9 for PB0/PB1 and some more. And AF14 for others.

ierturk:
Sorry, Thank you. This is my fault I did not noticed that for PB0 and PB1 needs to be set AF9 for LTDC. Please revert the merge.

burrbull:
If you trying to decrease boilerplate in user code  instead of `ltdc_pin` macro preferred way is using enums from `gpio::alt::ltdc`.
See https://github.com/stm32-rs/stm32f4xx-hal/blob/2f76eb926b1ee62d468a10671cda24d90fb96b95/src/spi.rs#L242
or more complex case like yours: https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/fsmc_lcd/pins.rs#L222

ierturk:
Ok. Thank you. I'll check them.

PR #710, opened 2023-12-07
move features to build.rs

burrbull:


Cargo.toml
```diff
@@ -338,91 +160,91 @@ opt-level = "s"
...
 
 [[example]]
 name = "dynamic-gpio"
-required-features = ["device-selected"]
+required-features = ["stm32f411"]
 
 [[example]]
 name = "f413disco-lcd-ferris"
-required-features = ["gpiod", "gpioe", "gpiof", "gpiog", "fsmc", "fsmc_lcd"] # stm32f413
+required-features = ["stm32f413", "fsmc_lcd"]
```
therealprof, line 195: I was kind of hoping we would go the other way around and make them less MCU model dependent? Not really a big fan of declaring that this only works on stm32f413... Same for the others of course.

burrbull: Good point. But I don't have good answer.
Do you have ideas?

therealprof: Not really, except that this is going in the wrong direction. Not quite sure how we can accomplish more genericity with the build.rs idea.

Cargo.toml
```diff
@@ -484,69 +186,16 @@ fsmc_lcd = ["dep:display-interface", "dep:display-interface-04"]
...
-tim14 = []
-usart3 = []
-uart4 = []
-uart5 = []
-uart7 = []
-uart8 = []
-uart9 = []
-uart10 = []
+# Private features for testing only
+_pac_fsmc = []
```
burrbull, line 190: @therealprof 
I could not find any other solution for examples except fictive features.

src/i2s.rs
```diff
@@ -243,37 +243,31 @@ macro_rules! i2s {
 // Each one has to be split into two declarations because the F412, F413, F423, and F446
```
berkus, line 243: Comment mentions 423 but I don't see it in cfg checks below at all...

PR #717, opened 2023-12-20
update to `heapless=0.8`

rursprung:


rursprung:
note: i explicitly didn't add this to the changelog as it's just a dev-dependency and thus has no impact on consumers (= readers of the changelog)

PR #720, opened 2023-12-29
swap embedded-hals

burrbull:


src/gpio/hal_02.rs
```diff
@@ -39,16 +37,6 @@ impl<const P: char, const N: u8, MODE> StatefulOutputPin for Pin<P, N, Output<MO
     }
 }
 
-impl<const P: char, const N: u8, MODE> ToggleableOutputPin for Pin<P, N, Output<MODE>> {
```
rursprung, line 42: why are you removing this for e-h 0.2? i don't think that you should remove these traits yet (all 0.2 traits should be removed together at some point)

src/prelude.rs
```diff
@@ -36,15 +36,15 @@
 //! ```
 //! use stm32f4xx_hal::prelude::*;
```
rursprung, line 37: if you want to make e-h 1.0 the default, shouldn't this then also be reflected in the prelude which should then export the e-h 1.0 traits instead of the e-h 0.2 traits? though this would be a breaking change for consumers relying on e-h 0.2 APIs which were re-exported here (as most of them changed between 0.2 and 1.0 or were removed altogether).

if you don't do the change now it'd at least be good to have an understanding for when this is planned to happen.

rursprung (commented):
FYI: i just noticed that the UART-related e-h 1.0 traits (which aren't in e-h but in `embedded-io`) are not yet implemented here. i've raised #721 for this

src/gpio/hal_02.rs
burrbull: thanks for catch. Fixed

burrbull:
> FYI: i just noticed that the UART-related e-h 1.0 traits (which aren't in e-h but in `embedded-io`) are not yet implemented here. i've raised #721 for this

Yes. There is a lot of traits not implemented yet.
Please open independent PR, if you have a draft.

therealprof (approved):
LGTM, thanks.

PR #724, opened 2024-01-10
add VS Code & JetBrains IDEA (RustRover, CLion, etc.) to `.gitignore`

rursprung:


rursprung:
note that i intentionally didn't add this to the changelog, it's irrelevant for consumers of the crate. it's just helpful when working with the repository with an IDE to avoid constantly seeing an untracked file being reported

burrbull:
In this logic `.lapce` should also be there.

rursprung:
> In this logic `.lapce` should also be there.

done. i wasn't aware of this IDE

burrbull:
`no changelog` label does not work for this repo. Don't know why. Need to fix.

PR #726, opened 2024-01-11
nb::serial

burrbull:


therealprof (approved):
LGTM, thanks.

therealprof:
NB: An example would have been nice.

PR #733, opened 2024-01-25
Sdio-host: update to 0.9.0, add AddressMode enum

skibon02:
Bump sdio-host to 0.9.0, which includes new voltabe_range format for send_op_conditon, generic arguments for registers (SD, EMMC).

Significantly, the SdioPeripheral::get_capacity() method has been replaced with SdioPeripheral::get_address_mode(). This change provides a more intuitive and accurate representation for eMMC cards, which unlike SD cards, do not have a distinction between standard and high capacity variants and uniformly use block addressing. The previous implementation relied on the high capacity bit in the OCR register, which is not applicable to eMMC cards. This bit has been removed in the sdio-host library to reflect the correct specifications for eMMC cards.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,5 +1,9 @@
 on:
-  pull_request_target:
+  pull_request:
+    # Run on labeled/unlabeled in addition to defaults to detect
+    # adding/removing skip-changelog labels.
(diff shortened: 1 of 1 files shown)


burrbull:
Looks good. Is it tested for both SD and eMMC? Update changelog and rustfmt, please.
Thanks.

PR #737, opened 2024-02-02
Prevent starting a new I2C transmission before previous stop finishes.

ZhiyaoMa98:
When a stop condition is generated by the I2C DMA ([example](https://github.com/stm32-rs/stm32f4xx-hal/blob/533797526ec093d5c04bad02171cba856e30d20d/src/i2c/dma.rs#L632)), the [code](https://github.com/stm32-rs/stm32f4xx-hal/blob/533797526ec093d5c04bad02171cba856e30d20d/src/i2c/dma.rs#L456-L458) does not wait for the finish of the stop condition. However, the [comments](https://github.com/stm32-rs/stm32f4xx-hal/blob/533797526ec093d5c04bad02171cba856e30d20d/src/i2c/dma.rs#L631) **falsely** indicate that there will be a spin-wait.

The I2C interface can get stuck if a new I2C transmission attempts to start while the previous stop condition is being generated. The new transaction will be inadvertently terminated by the previous stop condition. Subsequently, the I2C interface gets stuck. The following figure shows that a previous stop condition being generated can immediately terminate a new start condition, and after that the I2C interface is stuck.

image

Since ISRs should finish as quickly as possible, the fix should not let ISRs spin-wait in the interrupt handler until the stop condition finishes. Instead, the spin-wait should be placed before starting a new transmission.

The patch also removes incorrect comments that falsely indicate that the code will spin-wait until the stop condition finishes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - Fix transmission termination in I2C master DMA read [#736]
+ - Prevent starting a new I2C transmission before previous stop finishes [#737]
 
 ## [v0.20.0] - 2024-01-14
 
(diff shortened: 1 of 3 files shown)


burrbull:
Looks reasonable for me.

cc @therealprof What do you think?

therealprof (approved):
LGTM, too.

PR #738, opened 2024-02-05
Implement Serial DMA abstraction.

ZhiyaoMa98:
The implementation follows existing DMA abstraction for I2C interface.

The implementation has been tested on STM32F407-Discovery board with the code in `examples/serial-dma.rs`.

### DIFF
--- a/src/serial/dma.rs
+++ b/src/serial/dma.rs
@@ -1,0 +1,658 @@
+use core::{marker::PhantomData, mem::transmute, ops::Deref};
+
+use super::{Instance, RegisterBlockImpl, Serial};
+use crate::dma::{
+    config::DmaConfig,
+    traits::{Channel, DMASet, DmaFlagExt, PeriAddress, Stream, StreamISR},
+    ChannelX, MemoryToPeripheral, PeripheralToMemory, Transfer,
+};
+use crate::ReadFlags;
+
+use nb;
+
+#[non_exhaustive]
+pub enum Error {
+    SerialError(super::Error),
+    TransferError,
+}
+
+/// Tag for TX/RX channel that a corresponding channel should not be used in DMA mode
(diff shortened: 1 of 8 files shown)


src/serial/dma.rs
```diff
@@ -0,0 +1,711 @@
...
+    /// # Warning
+    /// `callback` may be called before function returns value. It happens on errors in preparation stages.
+    unsafe fn read_dma(
+        &mut self,
+        buf: &mut [u8],
+        callback: Option<SerialCompleteCallback>,
+    ) -> nb::Result<(), super::Error>;
+}
+
+pub trait SerialWriteReadDMA {
```
burrbull, line 64: What use-case of WriteRead in with Serial?

Acciente717: To be honest, I don't have a current use case for this. I was basically following the existing implementation of I2C and made it to similarly work for serial.

However, I do think there can be use case for such communication. For example, when communicating with another chip connected over serial, the WriteRead can be used to support the paradigm of send query and read response from the external chip.

burrbull: I2C is synchronous interface. UART is asynchronous.  This paradigm will not work properly without synchronization.

Please, remove this part from PR.

ZhiyaoMa98: Thanks. `SerialWriteReadDMA` is removed.

burrbull (approved):

burrbull:
Sorry. One more. Do not use `merge master` in PR, please.
`rebase` it instead.

Acciente717:
No problem. Done.

Acciente717:
Ah sorry seems like I brought back the `SerialWriteReadDMA` trait. Let me remove it again.

burrbull (approved):

PR #742, opened 2024-02-09
I2C+DMA: Set LAST on last RX DMA transfer

X-yl:
https://www.st.com/resource/en/reference_manual/rm0402-stm32f412-advanced-armbased-32bit-mcus-stmicroelectronics.pdf 

See 24.6.2:

> 
> LAST: DMA last transfer
>     0: Next DMA EOT is not the last transfer
>     1: Next DMA EOT is the last transfer
>     Note: This bit is used in master receiver mode to permit the
>     generation of a NACK on the last received data.
> 

This solves issues with peripherals like the BMP581 which will keep the i2c bus busy until it receives a NACK.


burrbull:
Is this related to https://github.com/stm32-rs/stm32f4xx-hal/pull/736 ?

X-yl:
Well this is embarrassing! I totally missed that PR, yes it's the same thing.

PR #754, opened 2024-05-14
fix: complementary output polarity

Orange-Murker:
`set_complementary_polarity` was setting the polarity for the for the primary output instead of the complementary output


burrbull (approved):

burrbull:
update changelog, please

PR #759, opened 2024-06-10
stm32-fmc

burrbull:


examples/fmc-sdram.rs
```diff
@@ -0,0 +1,127 @@
+//! This example shows how to use the FMC controller on the STM32F7 to communicate with an
+//! off-chip SDRAM memory device. The `stm32-fmc` crate does the majority of the work, and
+//! after initialization the SDRAM is memory mapped to the STM32F7 address space.
+//!
+//! This example was tested on the STM32F746G Discovery Board. The board has an IS42S32400F-6BL
```
therealprof, line 5: It would be nice if the example actually mentions a real STM32F4xx usecase, it's hard to see how to apply this to a F4 board.

tegimeki: I have been using the branch from this PR on an STM32F469I-DISC board, example [here](https://gitlab.com/mfairman/stm32f469i-disc/-/blob/mfairman/sdram/examples/fmc_sdram_test.rs) which does a memory test.

However, that code is blocked on [stm32-fmc](https://github.com/stm32-rs/stm32-fmc) having the SDRAM chip for that board added ([PR here](https://github.com/stm32-rs/stm32-fmc/pull/14)), so I'm just using local clones of these two crate repos for now... perhaps once `stm32-fmc` changes are merged we can combine examples?

burrbull: Does it work as expected?

burrbull: Could you correct PR example to make it more F4?

tegimeki: It works 100%, but depends on the other chip being added to stm32-fmc; once that is available I'd be happy to bring my memory test here (or you can feel free to take it and adapt, either way).

burrbull: > I have been using the branch from this PR on an STM32F469I-DISC board, example [here](https://gitlab.com/mfairman/stm32f469i-disc/-/blob/mfairman/sdram/examples/fmc_sdram_test.rs) which does a memory test.

Replaced example with yours (disabled for now).
As we do not depend on https://github.com/stm32-rs/stm32-fmc/pull/14 directly I'm going to merge this PR.
So we will need just enable example when `stm32-fmc` is released.

PR #768, opened 2024-06-15
UART::steal & Deref

burrbull:


therealprof (commented):
It's a bit weird to have separate `{rx|tx}_peri_address` methods when there really is only one data register which is used for both. Even more weird that we do it for UART and I2C but not SPI... Any reason for that?

burrbull:
> It's a bit weird to have separate `{rx|tx}_peri_address` methods when there really is only one data register which is used for both. Even more weird that we do it for UART and I2C but not SPI... Any reason for that?

I'm still sometimes working on possibility to integrate other families (which have separate registers for RX and TX), so I'm trying to eliminate inconsistencies to simplify code sharing.
Added for SPI too.

therealprof (approved):
LGTM

PR #772, opened 2024-07-16
I2C 10-bit address support

burrbull:
Needs tests with 10-bit I2C address devices!!!


burrbull:
Merging this as other PRs depend on it.

PR #786, opened 2024-08-02
dsi: add support for DSI host peripheral

tegimeki:
Adapted from the stm32h7xx-hal implementation, changed to use register accessor functions and use the RCC enable/reset mechanisms which are different in the stm32f4xx-hal.

Changes the `LtdcPins` argument to the `DisplayController` to be an `Option`, as they aren't used (and aren't needed for DSI interfaces).

Tested on STM32F469I-DISC board, in [this project](https://gitlab.com/mfairman/stm32f469i-disc/-/compare/mfairman%2Fsdram...mfairman%2Fdsi-lcd?from_project_id=60414443&straight=false) (pending various crate updates before that BSP is directly usable)

### DIFF
--- a/src/dsi.rs
+++ b/src/dsi.rs
@@ -1,0 +1,695 @@
+//! Display Serial Interface
+//!
+//! Interface with MIPI D-PHY
+
+use crate::ltdc::DisplayConfig;
+use crate::rcc::{Clocks, Enable};
+use crate::{pac::DSI, time::Hertz};
+use core::cmp::{max, min};
+use embedded_display_controller::dsi::{DsiHostCtrlIo, DsiReadCommand, DsiWriteCommand};
+
+const DSI_TIMEOUT_MS: usize = 100;
+
+pub struct DsiHost {
+    dsi: DSI,
+    channel: DsiChannel,
+    cycles_1ms: u32,
+}
+
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum DsiChannel {
+    Ch0 = 0b00,
+    Ch1 = 0b01,
+    Ch2 = 0b10,
+    Ch3 = 0b11,
+}
+
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Error {
+    RegTimeout,
+    PllTimeout,
+    BufferIsToBig,
+    WriteTimeout,
+    ReadTimeout,
+    ReadError,
+    FifoTimeout,
+    WrongId,
+}
+
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum DsiMode {
+    Video { mode: DsiVideoMode },
+    AdaptedCommand { tear_effect: Option<TearEffectMode> },
+}
+
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum DsiVideoMode {
+    NonBurstWithSyncPulses = 0b00,
+    NonBurstWithSyncEvents = 0b01,
+    Burst = 0b10,
+}
+
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct TearEffectMode {
+    pub source: TearEffectSource,
+    pub auto_refresh: bool,
+}
+
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum TearEffectSource {
+    DsiLink,
+    ExternalPin,
+}
+
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum DsiInterrupts {
+    None,
+    All,
+}
+
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum DsiCmdModeTransmissionKind {
+    AllInHighSpeed,
+    AllInLowPower,
+}
+
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct DsiPhyTimers {
+    pub dataline_hs2lp: u8,
+    pub dataline_lp2hs: u8,
+    pub clock_hs2lp: u16,
+    pub clock_lp2hs: u16,
+    pub dataline_max_read_time: u16,
+    pub stop_wait_time: u8,
+}
+
+pub struct DsiRefreshHandle {
+    dsi: DSI,
+    // refresh_request: *mut bool,
+}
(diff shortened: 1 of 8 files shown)


src/dsi.rs
```diff
@@ -0,0 +1,693 @@
...
+//!
+//! Interface with MIPI D-PHY
+
+use crate::ltdc::DisplayConfig;
+use crate::rcc::{Clocks, Enable};
+use crate::{pac::DSI, time::Hertz};
+use core::cmp::{max, min};
+use embedded_display_controller::dsi::{DsiHostCtrlIo, DsiReadCommand, DsiWriteCommand};
+//use embedded_display_controller::DisplayConfiguration;
+#[cfg(feature = "log")]
```
burrbull, line 11: There is no `log` feature in f4xx-hal. Unless you want to add it.

src/dsi.rs
```diff
@@ -0,0 +1,693 @@
...
+        self.dsi
+            .cltcr()
+            .modify(|_, w| unsafe { w.hs2lp_time().bits(max_time).lp2hs_time().bits(max_time) });
+        self.dsi.dltcr().modify(|_, w| unsafe {
+            w.mrd_time()
+                .bits(phy_timers.dataline_max_read_time)
+                .hs2lp_time()
+                .bits(phy_timers.dataline_hs2lp)
+                .lp2hs_time()
+                .bits(phy_timers.dataline_lp2hs)
```
burrbull, line 472: More compact and `rustfmt` friendly style to write the same
```suggestion
            w.mrd_time().bits(phy_timers.dataline_max_read_time);
            w.hs2lp_time().bits(phy_timers.dataline_hs2lp);
            w.lp2hs_time().bits(phy_timers.dataline_lp2hs)
```

src/dsi.rs
tegimeki: True; this came from the h7xx-hal so I can look at adding it in here, or will remove the code using it if that gets out of scope.

src/dsi.rs
burrbull: Same for other similar field chains.

src/dsi.rs
tegimeki: I removed the logging code, along with some other leftover debug things I inadvertently left in.

It would be interesting to have an `stm-dsi` crate at some point, given the DSI block is the same across the f4/f7/h7 families (or close enough to `cfg`ify), and with that some of the other elements could be added.

src/dsi.rs
tegimeki: Done.  There are some other uses of chained writes in the original H7 code, left as-is.

tegimeki:
> > Changes the `LtdcPins` argument to the `DisplayController` to be an `Option`, as they aren't used (and aren't needed for DSI interfaces).
> 
> Not sure what is better. `Option` or additional constructor.

I wasn't sure either, so will follow any recommendations.  Since the parameter isn't used it could be removed altogether; the purpose of the pin declarations appears to be for setting the ALT mode, which gets done by the calling code anyway.  But if there are other intentions for this parameter we could make a separate constructor for the pin-less DSI case.

src/dsi.rs
```diff
@@ -0,0 +1,672 @@
...
+#[derive(Copy, Clone)]
+pub enum DsiChannel {
+    Ch0 = 0b00,
+    Ch1 = 0b01,
+    Ch2 = 0b10,
+    Ch3 = 0b11,
+}
+
+#[derive(Debug)]
+pub enum Error {
```
burrbull, line 28: All enums should implement these traits:
```suggestion
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum Error {
```

src/dsi.rs
```diff
@@ -0,0 +1,672 @@
...
+    // refresh_request: *mut bool,
+}
+// unsafe impl Send for DsiRefreshHandle {}
+
+pub enum LaneCount {
+    SingleLane = 0b00,
+    DoubleLane = 0b01,
+}
+
+pub struct DsiPllConfig {
```
burrbull, line 91: Same traits for simple structs.

src/dsi.rs
burrbull: For note: I'm experimenting with merge f4 and f7 hals as most close to each other: https://github.com/stm32-rs/stm32f4xx-hal/pull/667 . But I need testers as I do not have F7 devices.

src/dsi.rs
tegimeki: Added, and tested output for a few:
```
dsi_config: DsiConfig { mode: Video { mode: Burst }, lane_count: DoubleLane, channel: Ch0, hse_freq: Rate { raw: 8000000 }, ltdc_freq: Rate { raw: 27429000 }, interrupts: None, color_coding_host: TwentyFourBits, color_coding_wrapper: TwentyFourBits, lp_size: 4, vlp_size: 4 }
dsi_pll_config: DsiPllConfig { ndiv: 125, idf: 2, odf: 0, eckdiv: 4 }
```

burrbull:
LGTM. Do you have an example to add?

src/dsi.rs
tegimeki: That is interesting, and quite an endeavor... in such cases would a new crate be warranted, if only to have a name which reflects the combined support?  I think we'd want any future F7 and F4 HAL work to go in only one place.

I don't have any F7 devices currently but will be happy to test things if I wind up getting one.

tegimeki:
> LGTM. Do you have an example to add?

I have an example in my [BSP project](https://gitlab.com/mfairman/stm32f469i-disc/-/compare/mfairman%2Fsdram...mfairman%2Fdsi-lcd?from_project_id=60414443&straight=false) but since that board requires SDRAM for framebuffer, it depends on https://github.com/stm32-rs/stm32-fmc/pull/14 and https://github.com/stm32-rs/stm32f4xx-hal/pull/759 in order to work.

I could add an example which configures the DSI peripheral on the STM32F469I-DISCO board and puts it in pattern test mode, as this does not require SDRAM and I might even be able to pare-down or eliminate the LTDC in this case.  I will try this out and if that would be interesting I can add it to this pull request.

burrbull:
> I could add an example which configures the DSI peripheral on the STM32F469I-DISCO board and puts it in pattern test mode, as this does not require SDRAM and I might even be able to pare-down or eliminate the LTDC in this case. I will try this out and if that would be interesting I can add it to this pull request.

Ok. Let's merge this PR and you open a new one with example when it will be ready.

burrbull (approved):

joelsa:
> I have an example in my [BSP project](https://gitlab.com/mfairman/stm32f469i-disc/-/compare/mfairman%2Fsdram...mfairman%2Fdsi-lcd?from_project_id=60414443&straight=false) but since that board requires SDRAM for framebuffer, it depends on https://github.com/stm32-rs/stm32-fmc/pull/14 and https://github.com/stm32-rs/stm32f4xx-hal/pull/759 in order to work.
> 
> I could add an example which configures the DSI peripheral on the STM32F469I-DISCO board and puts it in pattern test mode, as this does not require SDRAM and I might even be able to pare-down or eliminate the LTDC in this case.  I will try this out and if that would be interesting I can add it to this pull request.

You can put an image in the internal flash, that way you also don't need the SDRAM.

It's what i did in this PR with an Example: https://github.com/embassy-rs/embassy/pull/2976

tegimeki:
> You can put an image in the internal flash, that way you also don't need the SDRAM.
> 
> It's what i did in this PR with an Example: [embassy-rs/embassy#2976](https://github.com/embassy-rs/embassy/pull/2976)

Good point and nice example @joelsa!

Though ultimately SDRAM is needed for rendering dynamically, I did make a simple HAL test which uses the built-in pattern generator in #789.  Hopefully your stm32-fmc change will get in soon 😄

PR #791, opened 2024-08-04
move PWM pins connecting after inialization

burrbull:
The main idea is from https://github.com/stm32-rs/stm32f3xx-hal/pull/34 but with several changes.

Always split PWM on manager and channels.

The main disadvantage is removed `embedded_hal_02::Pwm` as no more `Pwm` struct.

cc @apollolabsdev @kossnikita

### DIFF
--- a/src/timer/pwm.rs
+++ b/src/timer/pwm.rs
@@ -7,397 +7,85 @@
 //!
 //! Second one is based on [`Timer`] with dynamic internally calculated prescaler and require [`fugit::Hertz`] to set period.
 //!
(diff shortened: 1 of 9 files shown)


src/timer/pwm.rs
```diff
@@ -7,682 +7,409 @@
...
-    }
+        // Trigger update event to load the registers
+        self.tim.trigger_update();
 
-    /// Set PWM channel duty cycle
-    #[inline]
-    pub fn set_duty(&mut self, duty: u16) {
-        TIM::set_cc_value(C, duty as u32)
-    }
+        self.tim.start_pwm();
```
burrbull, line 101: Should it be stopped here? Add `pub fn start/stop` ?

PR #794, opened 2024-08-14
I2C read address fix

jagenheim:
I have not verified the change for the 10-bit addresses, but it is likely to suffer the same problem.


burrbull (approved):

burrbull:
Thanks a lot.

burrbull:
most likely too in line 351

PR #807, opened 2024-10-05
fix defmt dependency enabling

### LINKED ISSUE #806: 0.22 pulls in defmt even if feature flag not activated
... since the `stm32f4` dependency is declared with the `defmt` flag unconditionally.

It should be activated by the `defmt` feature instead.

burrbull:
Fixes #806


birkenfeld:
Works for me, thanks a lot!

PR #813, opened 2024-10-12
remove NoPin, use Option

burrbull:


therealprof (approved):
I like it. I'm wondering how we came up with the `NoPin` concept to begin with.

burrbull:
> I like it. I'm wondering how we came up with the `NoPin` concept to begin with.

1. There were `NoSck` implementing `PinSck` trait, `NoMiso`, etc. 
2. Then I added common `NoPin` when generic `Pin` was introduced.
3. When alternate pin enums were added I was forced to add `NoPin` variant to that enums just because I could not find better solution.

`Option` based way is also not ideal as `None` can't infer its type. That is why `SPIx::NoMiso` and relatives were added.
But it is definitely better than `NoPin` enum variant.

PR #829, opened 2024-12-28
Use write instead of modify to clear I2C::SR flags

burrbull:


burrbull:
@pdgilbert I'm not sure but this looks reasonable for me.

PR #830, opened 2025-01-01
Added support for Input Capture timer mode for frequency and pulse duration measurement

faunel:
Added support for the Input Capture timer mode.

Why this is needed:

The Input Capture mode allows measuring signal frequency or pulse duration. This is especially useful for tasks that require precise frequency measurement, such as monitoring fan RPM, working with sensors, or analyzing signals. Additionally, a single timer can use up to four channels, enabling the measurement of multiple signals simultaneously.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -26,6 +26,10 @@ pub use pwm::*;
 pub mod pwm_input;
 #[cfg(not(feature = "gpio-f410"))]
 pub use pwm_input::PwmInput;
+#[cfg(not(feature = "gpio-f410"))]
+pub mod capture_compare;
+#[cfg(not(feature = "gpio-f410"))]
+pub use capture_compare::*;
 #[cfg(feature = "rtic1")]
 pub mod monotonic;
 #[cfg(feature = "rtic1")]
@@ -87,6 +91,7 @@ pub const C4: u8 = 3;
 pub enum Polarity {
     ActiveHigh,
     ActiveLow,
+    ActiveBoth,
 }
 
 /// Output Idle state
@@ -306,6 +311,18 @@ pub enum Ocm {
     PwmMode2 = 7,
 }
 
+/// Capture/Compare mode
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(u8)]
+pub enum Ccm {
+    // Todo Output compare (not tested)
+    // OutputCompare = 0,
+    InputCapture = 1,
+    InvChannelInputCapture = 2,
+    TriggerInputCapture = 3,
+}
+
 // Center-aligned mode selection
 pub use pac::tim1::cr1::CMS as CenterAlignedMode;
 
@@ -320,7 +337,7 @@ pub type CCR4<T> = CCR<T, 3>;
 pub struct DMAR<T>(T);
 
 mod sealed {
-    use super::{BitFlags, CenterAlignedMode, Event, Flag, IdleState, Ocm, Polarity};
+    use super::{BitFlags, Ccm, CenterAlignedMode, Event, Flag, IdleState, Ocm, Polarity};
     pub trait General {
         type Width: Into<u32> + From<u16>;
         fn max_auto_reload() -> u32;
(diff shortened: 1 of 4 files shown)


src/timer.rs
```diff
@@ -359,6 +426,14 @@ mod sealed {
...
     }
 
+    pub trait WithCcCommon: General {
+        const CC_CH_NUMBER: u8;
+        const CC_COMP_CH_NUMBER: u8;
+        fn read_cc_value(channel: u8) -> u32;
+        fn enable_channel(channel: u8, b: bool);
+        fn set_channel_polarity(channel: u8, p: Polarity);
+    }
+
```
burrbull, line 436: I do not see ant sense in additional copy of `WithPwmCommon` trait.
It is better to just rename `WithPwmCommon` to something more common.

src/timer.rs
```diff
@@ -411,6 +501,26 @@ macro_rules! split {
     };
 }
 
+macro_rules! split_cc {
```
burrbull, line 504: This macro can also be joined with previous as it is implemented for same timers.

faunel:
Yes, I agree. That was code duplication.
I refactored it and added comments.

src/timer/capture.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub enum CaptureLines<P> {
+    One(P),
+    Two(P, P),
+    Three(P, P, P),
+    Four(P, P, P, P),
+}
```
burrbull, line 116: This was done to allow connecting several pins on same timer channel (although as for me I've never used it). Is it allowed in capture mode?

faunel: I don't understand how this is supposed to work.
I took a lot of code from pwm.rs as a basis. So if it worked there, it should also work in input capture mode.
If you tell me how I can test it, I'll try to do so.

faunel: If this is what you mean, then yes, it works:

```rust
let (mut tim5, (ch1, ..)) = Timer::new(dp.TIM3, &clocks).capture_hz(48.MHz());
let mut ch1 = ch1.with(gpioa.pa6).with(gpiob.pb4.into_alternate());
```

faunel: I tested further and got unexpected results.

The pins are as follows:  
TIM3 channel1 = PA6 original  
TIM3 channel1 = PB4 remap  

TIM3 channel2 = PA7 original  
TIM3 channel2 = PB5 remap  

### Using channel 1  
In this case, only PB4 is active:  
```rust
let (mut tim5, (ch1, _ch2, ..)) = Timer::new(dp.TIM3, &clocks).capture_hz(48.kHz());
ch1.with(gpiob.pb4.into_alternate()).with(gpioa.pa6.into_alternate());
```

If you swap them, the same happens: only PB4 is active:  
```rust
let (mut tim5, (ch1, _ch2, ..)) = Timer::new(dp.TIM3, &clocks).capture_hz(48.kHz());
let mut ch1 = ch1.with(gpioa.pa6.into_alternate()).with(gpiob.pb4.into_alternate());
```

### But when using channel 2, the result is different:  
Both pins PA7 and PB5 are active:  
```rust
let (mut tim5, (_ch1, ch2, ..)) = Timer::new(dp.TIM3, &clocks).capture_hz(48.kHz());
ch2.with(gpioa.pa7.into_alternate()).with(gpiob.pb5.into_alternate());
```

If you swap them, both pins are still active:  
```rust
let (mut tim5, (_ch1, ch2, ..)) = Timer::new(dp.TIM3, &clocks).capture_hz(48.kHz());
ch2.with(gpiob.pb5.into_alternate()).with(gpioa.pa7.into_alternate());
```

faunel: Personally, I don't see any benefit in this and can't come up with use cases for it.  
I would remove this entirely:  
```rust
pub enum CaptureLines<P> {
    One(P),
    Two(P, P),
    Three(P, P, P),
    Four(P, P, P, P),
}
```

Usually, it should be used like this:  
```rust
let timer = Timer::new(dp.TIM4, &clocks);
let (_, (ch1, ch2, ch3, ch4)) = timer.pwm_hz(25.kHz());
let mut ch1 = ch1.with(pwm4_ch1);
let mut ch2 = ch2.with(pwm4_ch2);
let mut ch3 = ch3.with(pwm4_ch3);
let mut ch4 = ch4.with(pwm4_ch4);
ch1.enable();
ch2.enable();
ch3.enable();
ch4.enable();
let tim_4 = (ch1, ch2, ch3, ch4);
```

burrbull: What about "complementary pins"?

faunel: In the input capture mode of the timer, complementary inputs do not exist.  
In this mode, both CC1P and CC1NP bits in the TIMx_CCER register  
are used to configure the input:  
- on the rising edge,  
- on the falling edge,  
- on both.


image

burrbull (approved):

burrbull:
@leonidv Could you test it with last changes?

leonidv:
I'm not  a senior embedded developer. For now, I am trying to write reading codes from an IR Remote (tsp2038) using capture compare. When I finish, will use this PR to make the same application.

burrbull (approved):

PR #836, opened 2025-05-25
use &mut RCC

burrbull:


burrbull:
This way is inspired by `stm32g0xx-hal`.

@therealprof What do you think. Are there pitfalls?

burrbull:
@therealprof review this PR too, please.

therealprof (approved):
Oof, big breaking changes.

X-yl:
Hi @burrbull , sorry for reviving this old PR, just wondering what is the rationale for this change? Is it required for soundness? It's a bit unfortunate for our large codebase which reuses a bunch of timers so now it is required to maintain critical sections in order to acquire `&mut Rcc` to reset timer peripherals.

burrbull:
> Is it required for soundness?

Mostly yes.

But you still can construct another `Rcc` if you are fine with `unsafe`. Something like:
```rust
let rcc2 = Rcc {
   clocks: rcc1.clocks,
   rcc: unsafe { pac::RCC::steal() },
};
```

PR #838, opened 2025-06-03
i2c: Implement embedded_hal::i2c::I2c for I2CMasterDMA

X-yl:
Rationale - Blocking access is already exposed via the I2CMasterDma::read + I2CMasterDma::write functions but the embedded_hal::i2c::I2c trait is not implemented meaning it does not integrate well with generic crates.

cc: @shonrb

### DIFF
--- a/src/i2c/dma.rs
+++ b/src/i2c/dma.rs
@@ -6,6 +6,8 @@ use crate::dma::{
     traits::{Channel, DMASet, DmaFlagExt, PeriAddress, Stream, StreamISR},
     ChannelX, MemoryToPeripheral, PeripheralToMemory, Transfer,
 };
+use crate::hal::i2c;
+use crate::i2c::dma::i2c::{ErrorType, Operation};
 use crate::ReadFlags;
 
 use nb;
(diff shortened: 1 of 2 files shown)


src/i2c/dma.rs
```diff
@@ -601,6 +603,36 @@ where
...
+    RX_TRANSFER: DMATransfer<&'static mut [u8]>,
+{
+    fn transaction(
+        &mut self,
+        addr: u8,
+        operations: &mut [Operation<'_>],
+    ) -> Result<(), Self::Error> {
+        for operation in operations {
+            match operation {
+                Operation::Read(dest) => self.hal_i2c.read(addr, dest),
```
burrbull, line 628: This annoying me. I've newer used `I2C+DMA`. Is this correct use call inner implementation? Does it actually use DMA?

burrbull: Could it be explained more widely in docs?

> same as the [methods here](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/i2c/dma.rs#L540-L581).

Yeah. These methods disturb me too.

PR #841, opened 2025-06-22
improve Ptr trait

burrbull:


burrbull:
@therealprof review, please.

PR #862, opened 2026-02-04
Serial IRQ example

mrwatts88:


burrbull:
Maybe name it `serial-echo-irq`?

mrwatts88:
> Maybe name it `serial-echo-irq`?

updated

burrbull (approved):

PR #843, opened 2025-06-22
touch and display example for STM32F469I-DISCO B08 board revision (NT35510 LCD controller)

Amperstrand:
## Summary

This PR adds NT35510 LCD controller support for the STM32F469I-DISCO B08 board revision, enabling both display and touch functionality.

I'm not much of a rust person, and my main goal was just trying to get it to work (which it now does). The code has been restructured into what I hope are clean, reusable commits with proper attribution to the original sources.

## What's Included

**Commits:**
1. `feat(ltdc)`: DSI constructor, framebuffer DrawTarget, and layer management APIs
2. `feat(display)`: NT35510 driver and F469DISCO board init helpers
3. `feat(examples)`: F469DISCO display examples with runtime panel autodetection
4. `fix(examples)`: Improved lcd-test panel detection and init sequence

**Attribution:**
- NT35510 init sequence based on [diybitcoinhardware/f469-disco](https://github.com/diybitcoinhardware/f469-disco) (MIT license)
- Display timing and patterns from STMicroelectronics STM32CubeF4 BSP (BSD-3-Clause)

## Hardware Tested

- ✅ STM32F469I-DISCO B08 revision (NT35510 LCD panel)
- ✅ Display initialization and test patterns
- ✅ Touch controller (FT6X06) calibration and input

## Questions

NT35510 could possibly be its own crate like otm8009a is, or alternatively it could live in `src/display/` as I've structured it here. Open to feedback on the best approach.

PR #863, opened 2026-02-25
feat(dsi,ltdc): implement missing DSI commands and add DSI-compatible LTDC constructor

Amperstrand:
## Summary

- Re-export DSI host trait and command enums from stm32f4xx-hal so external display drivers can depend only on the HAL.
- Implement DCS short and generic short write commands in the DSI host implementation, removing remaining todo!()s.
- Add a DSI-oriented LTDC constructor that does not touch PLLSAI and works with the DSI host pixel clock.

## Notes

This PR is the first of two extracted from #843 (NT35510 display+touch example). It contains only the generic DSI/LTDC transport improvements and no example or board-specific changes.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/dsi.rs
+++ b/src/dsi.rs
@@ -6,7 +6,7 @@ use crate::ltdc::DisplayConfig;
 use crate::rcc::{Enable, Rcc};
 use crate::{pac::DSI, time::Hertz};
 use core::cmp::{max, min};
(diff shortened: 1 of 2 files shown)


Amperstrand:
Consolidating into #843

PR #864, opened 2026-02-25
feat(example): add NT35510 LCD support with auto-detection for F469I-DISCO

Amperstrand:
## Summary

- Add a minimal NT35510 DSI LCD controller driver used by the F469I-DISCO LCD test example.
- Update the  example to auto-detect NT35510 vs OTM8009A at runtime (no HAL-level feature flags).
- Use the new  constructor so LTDC is driven from the DSI host PLL on STM32F469.

## Relationship to #843

This PR is the second of two smaller PRs extracted from #843 and focuses only on the F469I-DISCO NT35510/OTM8009A example. The generic DSI/LTDC transport changes now live in PR #863.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


Amperstrand:
Consolidating into #843

PR #874, opened 2026-05-11
Support Latest Embedded-io

ScottGibb:
This PR bumps the `embedded-io` dependency to use the latest version. Following the same pattern as the STM32F1 HAL PR: https://github.com/stm32-rs/stm32f1xx-hal/pull/565


ScottGibb:
@burrbull I think that should be us good to go! :)

burrbull (approved):

